PR #7, opened 2022-04-10
Fixed IsLocked call to None object, Improved readability

BreadDain:
Can't reproduce lockpicking issue reported by users anymore after the fix, but I also had difficulty narrowing down what is causing it so it still might still be present. In any case - one issue down.


BreadDain:
Update: found conditional bug that solves most problem (apparently I was too good in lockpicking to not cause it constantly)
1 issue still might persist, when closing menu without even trying to lockpick it might thing you've actually broken all lockpicks:
изображение

IHateMyKite:
Did you possibly take more than 25 seconds ? Because only for the result to change to 2 (that I see) is [here](https://github.com/BreadDain/UnforgivingDevices/blob/e177e6d2228c00a948e12253e8f1f96c153fc7cb/Scripts/Source/UD_CustomDevice_RenderScript.psc#L4360).

IHateMyKite (approved):
Nice work

IHateMyKite:
Try to fix the last issue. If it will be too hard, then merge and create a new issue entry.

BreadDain:
> Did you possibly take more than 25 seconds ? Because only for the result to change to 2 (that I see) is [here](https://github.com/BreadDain/UnforgivingDevices/blob/e177e6d2228c00a948e12253e8f1f96c153fc7cb/Scripts/Source/UD_CustomDevice_RenderScript.psc#L4360).

On opposite, I am picking very fast. I suspect that's something connected with NPC LoS lockpicking interruption when you break a lockpick, they might stop minigame is such fashion that timer doesn't reset and still counts up to 25. But that's a pure theory.

BreadDain:
Impossible to resolve without major change of design, Can't do merges, only re-request review.

IHateMyKite (approved):

PR #20, opened 2022-04-19
PermitAnal check fix

BreadDain:
Did a quick check on both types of belts, seems to work as intended now. Function might cause some slowdown so I marked it with appropriate comment.


IHateMyKite (approved):
Nice work. Also, the function GetWornRenderedDeviceByKeyword should be actually fast. It doesn't check inventory at all. It takes directly device by using NI slot. Edit: Let me know if you can/can't merge

BreadDain:
I am unable to merge pulls

PR #23, opened 2022-04-20
Enhancement of ActorIsFollower check with custom followers

BreadDain:


IHateMyKite (approved):
Nice. Thanks for info. Will do some research later.

PR #29, opened 2022-04-26
Fixes for helper animating message, changed black goo text to messagebox, fixed rogue armbinder message spam

BreadDain:
Split long messages in 2 (another option is messagebox which is annoying)
Fixed misplaced function, wearer/helper


IHateMyKite:
> I've also started some discussions here on git, thinking if I would be able to convert mod to use translation files instead of fixed phrases. Please tell me what do you think about it.

Did forget about the discussion page, will give it a look.

> Another thing I am thinking about is working on equip-random-restraint function, to replace it with more defined restraints from leveled lists, to avoid getting equipped undesired items such as Mask of Shame or colour mismatching rubber parts, yellow green and etc.

Seems like a great idea.

IHateMyKite (commented):
I think that in this case, it would be better to just show messagebox. The text is for RARE device ans so it will only show rarely. It will also make user better realize that rare device is being locked on. You can use this [function ](https://github.com/IHateMyKite/UnforgivingDevices/blob/40e8efb1f1a7bf77ee066ef8e535a865ce9fef53/Scripts/Source/UDCustomDeviceMain.psc#L3493). Also, if you will in future add debug.notification or rework the old one, try to use [this ](https://github.com/IHateMyKite/UnforgivingDevices/blob/40e8efb1f1a7bf77ee066ef8e535a865ce9fef53/Scripts/Source/UDCustomDeviceMain.psc#L242)function. This function takes level variable which will be used in future to filter messages. Idea is that there are too many notifications ,and so I'm adding levels to them. User then can select in MCM up to which level they wan't to see the notification. If level will be 1, it means it will always show (like in this case). 3 is the less important one.

BreadDain:
```
`    Another thing I am thinking about is working on equip-random-restraint function, to replace it with more defined restraints from leveled lists, to avoid getting equipped undesired items such as Mask of Shame or colour mismatching rubber parts, yellow green and etc.`

Seems like a great idea.
```
Didn't take too long to get a hard to solve issue - reading keywords from leveled items. Organizing everything by colour/material is simple enough, but then I need to run keyword check just to get necessary devices from the whole list, and this implies spawning every object beneath player as reference. Seems there is no way other than pre-caching everything in JSON format.

BreadDain:
Added elbowbinder fix per https://www.loverslab.com/topic/165197-unforgiving-devices/?do=findComment&comment=3737996
Not completely sure if I did it right or even fixed it, but cant reproduce the issue anymore

IHateMyKite (changes requested):
Solution is actually not caused by this. Issue is that it doesn't check how much condiliton levels were added. Max number is 4, then the device is destroyed. So fix would be somethink like this
`while (_total_durability_drain >= 100) && UD_condition < 4`. But you can keep the fix you added and just add the UD_condition < 4 (only for the while loop)

IHateMyKite (approved):
Nice work!

Issue #40, opened 2022-05-03
(Major) getArousalSkillMult(Actor akActor) function fails to work properly and is incompatible with different Aroused version due to their structure


BreadDain:
Description: OSL is not completely backward compatibile and puts all aroused effects into FE range, thus breaking UD skill gain calculation.

![изображение](https://user-images.githubusercontent.com/30041703/166465707-10c2fc1e-2e29-4461-a9ad-604be2c6d438.png)

Version of skyrim: SE
Version of mod: 2.0b10
Used patches: DD 5.1, DCL, LBS

BreadDain:
Solution 1: fetching Skill Use actor values and modifying skill gains by missing about if it is below 1.0
Solution 2: use of soft dependency and checking where OSL or regular SLA is installed. Pending use of code request from Subhuman
```
;thanks to Subhuman#6830 for ESPFE form check, compatible with LE
form function GetMeMyForm(int formNumber, string pluginName) global
    int theLO = Game.GetModByName(pluginName)
    if ((theLO == 255) || (theLO == 0)) ; 255 = not found, 0 = no skse
        Debug.Trace(pluginName + " not loaded or SKSE not found", 1)
        return none
    elseIf (theLO > 255) ; > 255 = ESL
        ; the first FIVE hex digits in an ESL are its address, so a formNumber exceeding 0xFFF or below 0x800 is invalid
        if ((Math.LogicalAnd(0xFFFFF000, formNumber) != 0) || (Math.LogicalAnd(0x00000800, formNumber) == 0))
            Debug.Messagebox("Plugin " + pluginName + " has FormIDs outside the range\nallocated for ESL plugins!: " + formNumber)
            Debug.Trace("ESL-flagged plugin " + pluginName + " contains invalid FormIDs: " + formNumber, 2)
            return none
        endIf
        ; getmodbyname reports an ESL as 256 higher than the game indexes it internally
        theLO -= 256
        return Game.GetFormEx(Math.LogicalOr(Math.LogicalOr(0xFE000000, Math.LeftShift(theLO, 12)), formNumber))
    else    ; regular ESL-free plugin
        return Game.GetFormEx(Math.LogicalOr(Math.LeftShift(theLO, 24), formNumber))
    endIf
endFunction

; Spell Property slaDesireSpell auto

Float Function getArousalSkillMult(Actor akActor)
	int i = 0
	form slaDesireSpell = none
	
[...]

PR #43, opened 2022-05-04
Desire spell compatibility enhancement with OSL

BreadDain:
Making mod able to recognize whether it is regular SLA or OSL is installed and fetching Lover's Desire spell accordingly.
изображение
**Warning:** SLADesireSpell property in UD_CustomDevice_Quest becomes undeclared, unused and erroneous in case of use of OSL, a subject for deletion.
https://github.com/IHateMyKite/UnforgivingDevices/issues/40


BreadDain:
Technically values should work, I didn't check an acutal value output, only that code is able to fetch the form and read it properly in both scenarios on SE.
Differences between 2 aroused plugins:

изображение

изображение

BreadDain:
Edit: magic effect order is different, closing and fixing for now

PR #44, opened 2022-05-04
Fix for getArousalSkillMult

BreadDain:
Rebuilt the function from the scratch since it wasn't working properly anyway, used soft dependent perk declaration instead of magic effect https://github.com/IHateMyKite/UnforgivingDevices/issues/40 https://github.com/IHateMyKite/UnforgivingDevices/pull/43
Quickly tested on OSL, verified form ids for SLAX and SLA 28

Code is prone to fail at many places and might cause issues, I tried to test it as much as possible but still not sure if it wouldn't fail somewhere.

Warning: SLADesireSpell property in UD_CustomDevice_Quest becomes undeclared, unused and erroneous in case of use of OSL, a subject for deletion.


IHateMyKite (approved):
Nice work. Only think that I want you to know is that there is already check for OSL install. It's [OSLArousedInstalled ](https://github.com/IHateMyKite/UnforgivingDevices/blob/e3f6e933b3ed9679d552b294314d65174cb39cc8/Scripts/Source/UnforgivingDevicesMain.psc#L121) and it's true when OSL is installed. You can try to use it to make the function faster.

BreadDain:
UDCustomDeviceMain has to be recompiled for the changes to apply, my version was conflicting with repository so I've overwritten it with source. Would it be a good idea for me to upload only psc in future?

IHateMyKite (approved):
Nice work

Issue #58, opened 2022-06-05
(Minor) Purified goo interaction


BreadDain:
Description: contrary to commit description, Purified goo re-uses Cursed goo functionality. In other words - both of them do the same thing, equip Punishing restraints.
Version of skyrim: SE
Version of mod: 2.0b11 [72eaa74](https://github.com/IHateMyKite/UnforgivingDevices/commit/72eaa746ea174fb8c34296348770aecfcced5e8e)

Issue #57, opened 2022-06-05
(Minor) Punisher piercings are not replacing existing device


BreadDain:
Description: consuming Cursed or Purified goo while having existing clitoral piercing equipped doesn't replace it with punishing version as expected. It only adds Punishing piercing to inventory without equipping it.
Version of skyrim: SE
Version of mod: 2.0b11 [72eaa74](https://github.com/IHateMyKite/UnforgivingDevices/commit/72eaa746ea174fb8c34296348770aecfcced5e8e)

Issue #54, opened 2022-05-26
(Minor) Abadon cursed straitjacket masturbation animation 


IHateMyKite:
[https://www.loverslab.com/topic/165197-unforgiving-devices/?do=findComment&comment=3760816](https://www.loverslab.com/topic/165197-unforgiving-devices/?do=findComment&comment=3760816)

PR #59, opened 2022-06-06
Fix for purified goo

BreadDain:
Fix for https://github.com/IHateMyKite/UnforgivingDevices/issues/58
Changed magic effect that Purified Goo uses to proper one.


BreadDain:
Fix for https://github.com/IHateMyKite/UnforgivingDevices/issues/57 and probably for https://github.com/IHateMyKite/UnforgivingDevices/issues/54 + additional rare cases where straitjacket fails to equip. Didn't see it for myself but saw someone complaining, in any case, changing keyword from Armbinder to Straitjacket in plugin and changing function from SwapDevice to LockDevice(force=true) shouldn't harm... probably? Did it on quick hand, probably good idea to check or keep in mind in later versions.

IHateMyKite:
Nice work. There is problem that I already modified the .esp file. I will have to merge it manually.

BreadDain:
> Nice work. There is problem that I already modified the .esp file. I will have to merge it manually.

It is easier to do manually. Cursed straitjacket's render object keywords and Purified Goo used magic effect (misplaced, probably due to copy)

PR #66, opened 2022-07-02
Fixed shield reequip, added torch support

BreadDain:
Per [report](https://www.loverslab.com/topic/165197-unforgiving-devices/?do=findComment&comment=3800114)
Changed the script so it checks equipped left hand instead of getting worn form, checking for Shield and Light form type, remove keyword check for redundancy, added Storage util value unset so shield is not getting requipped after animation if it was manually removed by player.

### DIFF
--- a/Scripts/Source/zadlibs_UDPatch.psc
+++ b/Scripts/Source/zadlibs_UDPatch.psc
@@ -827,8 +827,8 @@ Function EndThirdPersonAnimation(actor akActor, bool[] cameraState, bool permitR
 			return
 		EndIf
 		Form loc_shield = StorageUtil.GetFormValue(akActor,"UD_UnequippedShield",none)
-		StorageUtil.UnsetFormValue(akActor,"UD_UnequippedShield")
 		if loc_shield
+			StorageUtil.UnsetFormValue(akActor,"UD_UnequippedShield")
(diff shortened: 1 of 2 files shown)


Scripts/Source/zadlibs_UDPatch.psc
```diff
@@ -826,7 +826,8 @@ Function EndThirdPersonAnimation(actor akActor, bool[] cameraState, bool permitR
 			UDCDMain.Error("EndThirdPersonAnimation("+GetActorName(akActor)+") - Actor is not loaded (Or is otherwise invalid). Aborting.")
 			return
 		EndIf
-		Armor loc_shield = StorageUtil.GetFormValue(akActor,"UD_UnequippedShield",none) as Armor
+		Form loc_shield = StorageUtil.GetFormValue(akActor,"UD_UnequippedShield",none)
+		StorageUtil.UnsetFormValue(akActor,"UD_UnequippedShield")
```
IHateMyKite, line 830: Please, move the function inside loc_shield condition, so it doesn't unset the form if its not set.

IHateMyKite (changes requested):
Please, fix the 1 problem. Otherwise nice work.

IHateMyKite (approved):
Nice work

PR #68, opened 2022-07-06
Inflation plug fix

BreadDain:
Override for inflatable plugs to check inflation level on removal. Function is the same as in UD_CustomPlug_RenderScript except || getPlugInflateLevel() > 0 is added in 2 places.

Supposed to disallow player to remove fully or partially inflated plugs immediately.


IHateMyKite:
Nice work. Like the change.

PR #92, opened 2022-09-04
Using group animation when struggling out of DD with helper

iiw2012:
I've made changes to the animation system to make struggling more intimate when using NPC help. If the player tries to get out of the device using help of the follower then a group animation will be triggered (like in SexLab). Ideally should be selected the most suitable animation concidered worn device and active constraints on actors.

The following changes have been made to implement this feature:

1. All constraints from heavy DD are defined in a bit array (see **UDCustomDeviceMain::GetActorConstraints** function). Added new function **UD_AnimationManagerScript::GetStruggleAnimationsByKeyword2** to select a solo struggle animation. This function is actually a wrapper for the existing **GetStruggleAnimationsByKeyword** method, which is called considering a bit array mentioned above. This allows you to localize the entire animation selection process in **UD_AnimationManagerScript**.

2. Getting the list, selecting and launching animations for the actors are localized in the **UD_CustomDevice_RenderScript::_PickAndPlayStruggleAnimation** function.
   - The list of animations is built by the **UD_AnimationManagerScript::GetStruggleAnimationsByKeywordWithHelper** function. It scans the <DATA>/skse/plugins/StorageUtilData/UD/StruggleAnims/ folder for JSON files which contains animation definitions suitable for certain devices with certain actor and helper constraints. See function description for JSON file format.
   - Then function **UD_AnimationManagerScript::FastStartThirdPersonAnimationWithHelper** starts animations for both actors synchronously in time and space.

3. The **UDCustomDeviceMain::UpdatePlayerControl** function has also been changed, as it was necessary to disable the player activation control during animation playback (otherwise the player would start a conversation with the NPC when pressing the action key).

4. I've selected several dozen more or less suitable animations from different SLAL Animation Packs and created behavior and JSON files for them.
   - 5 animations from SexLab (by Leito).
   - 23 animations by BakaFactory.
   - 30 animations by Billy.
These animations cover almost all simple devices and a couple of heavy boundages (see UD Paired Struggle.xls). All needed HKX files are currently included in the branch for easier testing, but may be extracted into some kind of animation packages in the future.

5. Added "doc/UD Paired Struggle.xls" in which custom FNIS_List and JSON files are defined and could be generated via VBA script.

6. **This branch contains animation and behavior files for Skyrim SE, their LE version is in another**:
https://github.com/iiw2012/UnforgivingDevices/tree/feature-paired-struggle-le

I've opened this pull request to get feedback on whether I'm moving in the right direction with certain changes. Or whether it was necessary to make these changes at all.

Next I have in plans:
- Improving and fixing current changes.
- Reworking the solo struggling animation system using JSON files.
- Paying more attention to the disable player control system.
- Adding custom animations instead of those I borrowed from SLAL Packs. But I can't give any guarantees that they won't be downright ugly.
  - Creating orgasm ending animations as continuation of struggling (using sequence animations like in SexLab).
- Using "aggressive" and "lewd" attributes from JSON (from other mods probably).


Scripts/Source/UDCustomDeviceMain.psc
```diff
@@ -2626,4 +2615,57 @@ float Function FinishRecordTime2(string strObject = "",bool bReset = false,bool
...
+	If akActor.WornHasKeyword(libs.zad_DeviousCuffsFront)
+		result[3] = True
+	EndIf
+	If akActor.WornHasKeyword(libs.zad_DeviousArmbinder)
+	; TODO: check for linked cuffs
+		result[4] = True
+	EndIf
+	If akActor.WornHasKeyword(libs.zad_DeviousArmbinderElbow)
+	; TODO: check for linked cuffs
+		result[4] = True
```
IHateMyKite, line 2655: This is probably causing elbowbinders to use armbinder animations

IHateMyKite:
Hello, this is incredible. I tested it and it mainly works. But there are still some issue which I found

- Elbowbinders use armbinder animations (I think that is caused by [this](https://github.com/iiw2012/UnforgivingDevices/blob/6870a30e41b6e6b906ad289423431ae48e355aae/Scripts/Source/UDCustomDeviceMain.psc#L2655))
- There are some issues with player disable. Mainly, huds don't show up if actor have blindfold effect from DD (not the default one, but the one which prevents movement in 3rd person camera). I have reworked the disable in yet unmerged fix, so you can then copy it to your branch. I will let you know once it is done.

Otherwise, it's working nice. My only personal issue might be that some animation that make it look like helper is trying to help player in other way then struggling *wink*

Also, did you get permission for the animations or are they for now only available for testing?

iiw2012:
Thanks for the quick response! Glad you liked it!

My changes in the player disable code are pretty sloppy, so I was going to revise that part. And propably everything else.

Regarding the animations used.

I haven't contacted the authors yet about using their animations in this mod.

The animation files from Billyy's SLAL Pack and SexLab don't really had to be included into the mod as I use their relative paths to the original location in the behavior files. Therefore, the animation files can be distributed with the original SLAL Packs.

Animations by BakaFactory are another matter since the path to the files contains a space. And GenerateFNISForModder does not process spaces correctly, so I had to copy these animation files into the folder with my FNIS List.

In any case, I have plans to slowly replace animations with specially made ones. It all depends on how well I can do it.

And yes, sometimes helper gets too carried away. That's why I've added "lewd" and "aggressive" attribute.

IHateMyKite:
Hello iiw2012, 
I have tried the latest version of this PR. The struggle animation is really great and smooth. I'm glad that there is finally some struggle animation for petsuit. 
But I have some problems actually running the mod. Mainly, for some reason, the whole function `_PickAndPlayStruggleAnimation` is behaving very strangely. The `UDAM.GetStruggleAnimationsByKeyword2(UD_DeviceKeyword_Minor, _actorConstraints)` works correctly and returns the value, but for some reason, _StruggleAnimationArray is not edited. It almost looks like that this is some compiler issue as papyrus mention this issue as `Error: Mismatched types assigning to variable named "::temp880"`, which doesn't make sense, because I checked that returned value is of the correct type. Do you have the same issue? If no, how do you compile the scripts?

iiw2012:
Hello! Sorry for the late response.

I've met that error, and, as it seems to me, I fixed it. 
It is my mistake, I should've checked the uninitialized array for Length == 0, instead of None. See UD_CustomDevice_RenderScript for the latest changes.

Also I've removed Billyy's animation from repo. All that files should be in the latest Billyy's animation pack.

I still not sure how better use animations from existing animation packs. Probably I will add checks if certain files exists before using appropriate json file with definitions.
For example. If there is Devious Devices - Assets.esm loaded then it is safe to use animations defined in UDStruggle_DB_DD.json.

iiw2012:
And one more thing I almost forgot about. I've played around with the  code from SexLab which removes the high heels from actors during animation. The problem is that this code drags along a dependency on the NiOverride class from the RaceMenu. See _RemoveHeelEffect and _RestoreHeelEffect from UD_AnimationManagerScript.

I raised this point so that I don't forget to finally resolve the issue with this dependency when it comes time to close this thread. I can't take the responsibility for adding a new dependency :) Even if it would be soft, it still requires NiOverride to be present when compiling. So the final word would be after you.

IHateMyKite:
Nice animations. I tested both of them and they look great. The only issue I had was that the actors were realigned every time the animation changed (every 3s by default if the feature is not turned off in MCM). There is no reason to do that as they were aligned in previous animations. I fixed it by simply changing the call of FastStartThirdPersonAnimationWithHelper in _PickAndPlayStruggleAnimation to the following:
`UDAM.FastStartThirdPersonAnimationWithHelper(Wearer, _minigameHelper, _sStruggleAnim, _sStruggleAnimHelper, !bUseCache)` . Note the last argument.

Also, something I wanted to ask last time but I forget. Do you also intend to make some arousal/orgasm/edge animations for petsuit? As this is the last kind of animation missing from the petsuit. Adding it would at least make it behave as every other hand restrains. I like that you made the pair animation in a way that wearer is on the back. Its great to see the petsuit from different angle. Maybe you could try to make an animation where the wearer is sitting ? (something like [this](https://64.media.tumblr.com/b6e9685bd699ffe7f216e8f9013d8baf/f99d7ee3110dde12-42/s250x400/db10290026e7be2ef83252bc242b095a2b5303c9.jpg) (didn't found better image, so sorry about that))

I'm also very thankful for your additions to the mod. I would have never been able to pull something like this on my own. I was once trying to make animations with blender, and even though they were not really bad (but very simple, I just wanted to make simple struggle animation for nipple piercings), I still didn't know how to implement them into skyrim with FNIS. So it's great there is someone who understands animations.

iiw2012:
I've included your fix in the last commit, thanks!

I thought about new orgasm/edge animations and probably will do it. As far as I can see, orgasm animations are played through the DD framework, so I first need to figure out at what stage it would be best to inject new animations.

And I'm very glad that you liked my animations! :) Creating animation for Skyrim is a real pain, especially if it's not your specialty. The first month I just set up the environment and selected import/export options so that it would at least work in the game. Now it should go easer.

Animation in a sitting pose is done, apparently, this was in the air. :)

By the way, that image gave me the idea that catsuit could have modificator that turns it into a latex pet suit (like linked gloves or cuffs). If you want, you can use it :)

iiw2012:
I've made major changes to the project. Moved all animations that depend on the version of the game to separate repositories. This will make it easier to work with them. Some animations (from DD and ZaZ) remained in place, as they do not require generated hkx files to use here.

- For Skyrim SE animations see https://github.com/iiw2012/UDAnimationsSE
- For Skyrim LE animations see https://github.com/iiw2012/UDAnimationsLE

IHateMyKite:
Nice work. I like where this is going. I have tested 20c6787 and found that there is some issue with horny animations. Looking at the scripts, it is caused by DD function AnimSwitchKeyword, but the function is already edited here https://github.com/IHateMyKite/UnforgivingDevices/blob/446a64142f0bcdfb18f5cc335ad840a33a1dab3e/Scripts/Source/zadlibs_UDPatch.psc#L567 
It can be fixed by adding something like this on the top of the function.
```
    if akActor.wornHasKeyword(zad_DeviousPetSuit)
      return ""
    endif
```
Or it can be edited in different way, I will keep it to you.

Also, the part needs to be edited for horny animation to be randomly played
https://github.com/IHateMyKite/UnforgivingDevices/blob/446a64142f0bcdfb18f5cc335ad840a33a1dab3e/Scripts/Source/UD_OrchamsCheckScript_AME.psc#L328-L330

iiw2012:
Oh! I completely forgot about zadlibs_UDPatch! Now there is no need to wait for the next DD update. :) Thanks for the tips, I'll take a look.

IHateMyKite:
Looks like you are making good progress. Once you finish all the changes you want to implement, please request the review, so I can know that the PR can be merged. In the meantime, I will do some fast test to test current implementation and give you feedback.

iiw2012:
I've added animations in the last weeks, and only checking that the code did not slow down more than usual. I still have a couple of ideas on how to speed up functions, but I don't want to complicate things unnecessarily. So we can assume that the code is stable more or less, unless you have any suggestions on this matter. In the meantime, I'm going to test, clean up the code and make small optimisations.

As for animations, most of the combinations are covered by added variants, except for exotic ones (for example when both actors have heavy bondage). But in these cases, my attempts so far end up either in banality or my lack of imagination :). I will check the uncovered combinations and try to close them.

IHateMyKite:
You can for not omit the optimization. I'm already thinking of some ways to make the scripts faster. Just clean the code as you mentioned, and then I will do the review and merge. This way you can focus more on the animations, and I will maintain and optimize the scripts.

iiw2012:
So, some intermediate results and plans.

In most cases, the new animation system works as intended. Apart from the incomplete coverage of all possible struggle variant, there are a couple of (minor) flaws:
1. When choosing an animation with multiple variants, only the first one that matches is selected. To some extent, this is how it was intended, but it can be done better.
https://github.com/iiw2012/UnforgivingDevices/blob/e3a132fb9c3dfa966e7b3a3712c6ae2fdcafbf76/Scripts/Source/UD_AnimationManagerScript.psc#L285-L288
Sometimes several variants of one animation are suitable, but this implementation does not take this into account.

2. Some DD have multiple keywords, but only one is used when choosing an animation (for example, DeviousHarness has additional keywords for Chastity and Collar).

3. The GetActorConstraintsInt function call takes a very long time (about 110 ms).
https://github.com/iiw2012/UnforgivingDevices/blob/e3a132fb9c3dfa966e7b3a3712c6ae2fdcafbf76/Scripts/Source/UD_AnimationManagerScript.psc#L697
I've already made some improvements, but ideally the value of this function should only be calculated when equipment is changed, and the stored value should be used in other calls.

4. There are log messages that lead to CTD with a large number of animations found.

5. An elusive bug with a shield appearing in animations. In most cases the shield disappears automatically when the animation starts, but sometimes it remains. I can assume that this is due to the companion's outfit with a shield, which is forced to be applied in some conditions. If this happens, then no attempts to remove the shield from the code helps.

The 1st issue is very minor and can be postponed until the next version.
The 2nd issue can affect performance and may require big changes because keywords will have to be checked.
The 3rd issue may require a separate PR, because a quick check for various equipped items can be useful not only in animations.
The 4th issue I will investigate and fix in the next commit.
The 5th issue I will investigate further in background.

I've tested the performance of the changes and we have about 10x headroom in animation count before the delays become noticeable (> 500 ms).

I also started writing documentation to make it easier for those who wish to create addons with new animations. 
https://github.com/iiw2012/UDAnimationsSE/wiki

And I continue to add animations to the best of my ability.

IHateMyKite:
Tested the newest version with last SE animations on SE. Didn't found any issue with scripts, looks like everything is working as intended. Only issue I have regarding scripts is that the time it takes for animation to switch to different ones is too small. But I was one who made that so I will have to adjust that later. Most likely will add MCM so the time can be adjusted.

Regarding the struggle animations, I have more issues. Petsuit animations works without issue, didn't found any clipping. But there are some clipping with different hand restrains.
- Yokes strangely twists the arms. See pics. First one is with one of the animation from SE pack (dunno which one, but I tried most of them and all had the same issue). Second one is default idle animation
Animation with helper
Default idle animation
- Elbowbinder clipping. In this case, its possible that it is caused by follower having high heels.
Elbowbinder
- Armbinders have same/similar issue (sorry for shity picture, had bad luck and got all black devices)
Armbinder

Last thing I found is that while in petsuit animation with helper, the helper fingers somehow gets twisted back in the hand (like shortened)(again sorry for shity picture, this is the best one I found).
hands

I thing I will also test the LE animations to see if its not caused by SE.

IHateMyKite:
Tried it on LE, have same issues.

iiw2012:
Thanks for the test!

Most of my animations are 5 seconds long. I planned to write information about the length of the animation in json so that switching between them was smoother. But I postponed the implementation, because I do not use this option in the game. :)

For some reason, I attributed short fingers to poor rigging of my models, and did not check how it would look in the game. I'll have to fix everything :) Same for the yoke's animations.

Regarding the third picture. That's the animation for armbinder is played due to my mistake in the json file.

And the forth issue is purely because of my low skill back then. I'll remove or replace this animation.

iiw2012:
Compared to my previous report :)
1. Reduced animation start time, and the freed up milliseconds is spent on better camera and position control.
2. Hopefully the shield bug won't bother us anymore.
3. Animation work is redirected to UD_AnimationManager.
4. MCM menu with features for testing.

There are still a couple of issues that I do not know how best to solve:
1. DD Framework generates animations (horny and edging idles) that are different from those that occur in UD. Namely, they can be interrupted by a jump. Is it worth trying to pass them through UD with all the locks?
2. The actor remains equipped with a sheathed weapon and a quiver. At the moment they do not interfere with animations, but this may change in the future. Do they need to be hidden for the duration of the animation?

If anyone has suggestions on what to add or change (or remove), I will gladly accept.

The same goes for animation. Now it seems that all common cases are covered, but if anyone has suggestions, what else can be added or changed, I'm ready to try to implement it.

IHateMyKite:
There is some strange fuckery happening when testing the animations. Basically, once the minigame is called, it takes quite long time to start (around 3-4 secs), and then it starts to move player with NPC around the world in one direction. Disabling the animations solves, the issue, so source of issue is when paired animation is played. Tried to backtrack to working version, and given up on commit d6d6738. I will try to test even older versions, but I think that issue will present all the way to aa579fb (it was working before, as you can see from my report about the animations). I will try the version which worked before to see if I'm just not doing something wrong.

iiw2012:
Does this happen when running animations through both MCM and dialogs ([ ; ] menus)?
Have you updated the mod with animations? I have changed the json file format a couple of times.
Lately, I've tested the changes in Skyrim SE only and everything works. I'll try LE.

iiw2012:
Don't waste your time testing.
I have already found several bugs that I will deal with in the near future. And once I got the situation as you described.

IHateMyKite:
Nwm, mine issue :D. After installing the mod in MO2, it didn't enable the PR patch. Because of that, the VehicleMarkerForm was set to none, which resulted in this. It might be a good idea to add a check on game reload which will fetch the form from the file, so it can never be none. I was also thinking if it will not be easier to just make one marker and then just move it around, instead of creating new one on every new animation. I might give it a try and commit it.

IHateMyKite:
Nwm again, that would because there can be multiple animations at once. Anyway now that I have found the issue I have tested it and everything is working as intended. Will play with it for some time to do some in terrain testing.

IHateMyKite:
Few issues I found so far:

- Solo struggle animation doesn't work if actor wears hobble skirt
- Sometimes, the armbinder solo animation is played instead of yoke solo animation (it's quite rare, so I think there is only 1 animation which is defined wrong in the json)
- After animation ends, it is impossible to immediately open the tween menu. Either it takes a long time to end, or there is an issue with enabling player control. There was a similar issue before in main branch, but I fixed it. You can try to use the function DisableActor and EnableActor from UDCDmain

If you are ok with i, I will start slowly to work on this PR too, as I see it as current priority before doing other features. I will for now take look at the issue with actor disable.

iiw2012:
There may be issues with the blocking system, as I don't fully understand how it works in the mod. I will accept any help.

In the meantime, I'll check the json for errors. Also there are some problems with newly added position save/restore code. I'll investigate that.

iiw2012:
I just found out that Elbowtie is not what I imagined the last months :)

ScreenShot20

These devices don't even present in UD patch, but I'll make animation versions for them anyway.

IHateMyKite:
> I just found out that Elbowtie is not what I imagined the last months :)
> These devices don't even present in UD patch, but I'll make animation variants for them anyway. There is also no solo animation for the hobble skirt. Will have to add that too.

They should be in UD patch. I remember getting some a few times, and they worked. Only issue is that some of the animations looks quite junky (mainly the walking idle animation)

iiw2012:
> They should be in UD patch. I remember getting some a few times, and they worked

Yes, I was wrong. For some reason the UD menu does not open for these devices, so I decided that they are not in the patch. In any case, it does not apply to this PR.


And a quick note on the last commit. I'm using VBA to generate json files, so it doesn't make much sense to separate them (only to set different load condition). I did not expect that they would have to be edited manually. Sorry if you had to do this :)

Here's a brief (perhaps even too brief) view on how it looks like. https://github.com/iiw2012/UDAnimationsSE/wiki/Using-VBA-to-generate-JSON-file-with-animations-defs

Spreadsheet with animations and VBA scripts is here https://github.com/iiw2012/UnforgivingDevices/blob/feature-paired-struggle/doc/UD%20Paired%20Struggle.xls

I'm actively editing jsons right now. I'll apply your changes through spreadsheet if you don't mind.

IHateMyKite:
I understood it wrong then. I don't mind if you implement the changes.

IHateMyKite:
I think I have found one "issue" which might be good to be addressed. Every time the actor have high heels, they will get little corrected for animation. Issue is that in some animations the actor is not standing on heels (for example in animation UD_Pair_Struggle_Plug01) actors are in stomach or on knees, so lowering them makes animation only more mismatched. I think it might be good idea to add another flag to excel sheet, which if flagged will prevent the high heels check (in case of paired animation, there will need to be 2 flags for both wearer and helper). What is your opinion ?

iiw2012:
The main task of removing HH effect is so that the actors are on the same height during animation. If after correction actors are even more misaligned then there is an error somewhere. That happens sometimes, but I haven't looked in that direction yet.

If I understand you correctly, additional flags for animations will not improve anything here. For example, if you do not remove the HH effects when the actors are in a lying position, then they will be hovering above the ground. Each on its own height.

Animation alignment is greatly hindered by the different scales of races in the game. Or different proportions of models. But if we delve into these issues, we will have to recreate SexLab (or something even more complex) :)

IHateMyKite:
I think that I explained it wrong, what I meant is that actor will have height changed even if there is no need for it, because the actor is not standing. If I understand it correctly, all the HH check does is that it position actor little lower if they have HH. Lets take for example this pic, where one actor have Restrictive Boots which count as HH, and other doesn't. Neither of them should have their height adjusted because they are not standing
pic

iiw2012:
As far as I know HH effect is a positive Z-shift applied to actor's bone "NPC". And those bones must be aligned during any group animation regardless of actors' limbs positions.

As for the screenshot, its seems like actor with boots remains its HH effect and hovers above the ground. Plus, this particular animation may be played with issues. For some reason, the helper is bent at the torso and unable to properly grab onto the plug. Something related with the physics engine probably.

I've tested other animations in different poses (SexLab pack has several boots struggle animations in every positions) but didn't saw misalignment.

I admit that I may be missing something in this matter. For now I can implement a more accurate calculation of HH shift that could eliminate some potential issues.

iiw2012:
I've got a question about json changes in commit [0631f45](https://github.com/IHateMyKite/UnforgivingDevices/pull/92/commits/0631f450744b01710b3de764399bc3a076eb5a5c). Regarding ft_horny_*** animations.

I reverted the UD_OrgasmAnimation flag (at least that was the intention) in zab_Libs function in commit [b824606](https://github.com/IHateMyKite/UnforgivingDevices/pull/92/commits/b8246068a1880ab2b5cfab21cef1c75cb58fa4cf) to reduce json file and make it more consistent. So that any horny animation can be played during orgasm if the flag is set.

Could your latest changes be achievable if we extend the influence of this flag to all orgasm animations calls not only from DD Framework? Or you had other motives to change "horny", "edge", "orgasm" keywords for ft_horny_*** animatons?

IHateMyKite:
> I've got a question about json changes in commit [0631f45](https://github.com/IHateMyKite/UnforgivingDevices/pull/92/commits/0631f450744b01710b3de764399bc3a076eb5a5c). Regarding ft_horny_*** animations.
> 
> I reverted the UD_OrgasmAnimation flag (at least that was the intention) in zab_Libs function in commit [b824606](https://github.com/IHateMyKite/UnforgivingDevices/pull/92/commits/b8246068a1880ab2b5cfab21cef1c75cb58fa4cf) to reduce json file and make it more consistent. So that any horny animation can be played during orgasm if the flag is set.
> 
> Could your latest changes be achievable if we extend the influence of this flag to all orgasm animations calls not only from DD Framework? Or you had other motives to change "horny", "edge", "orgasm" keywords for ft_horny_*** animatons?

I see, I might have understand it wrong. In that case it looks like that it is either not working. It only plays orgasm animations, even when the extended list should be used. Will look at it in the papyrus.

IHateMyKite:
Hi,

I have question regarding the progress. Are there still some mayor changes planned, or are you just fine-tuning the things ? If it's the second case, I think it would be better to for now merge this PR, and later create new one. As I have said on LL. Once this is done, I will publish the existing MB as some kind of beta (similar to DD beta) on LL support thread. So you can keep working on improvements, while also get some feedback from other users on LL.

Also, sorry for implementing non-related things, I'm testing the animation system on quite long game session, so I could not go back to main branch, so I was implementing features here. This way, I can also test if the features will work on existing saves.

Lastly, I think that at this point, there are no more issues from my point of view (regarding the scripts). I think it works quite fast, but I'm using the Papyrus NG, so It's hard to say. It's still obviously slower than previous implementation, but not in such magnitude that it would make me think that it is slow.

Only thing which comes to my mind that the new corset struggle animation looks little funky (one where character use one hand on stomach, and other on back). The hand used on the back doesn't get correctly rotated, so it looks like the character have a twisted wrist. Otherwise, I have not found any other issue.

iiw2012:
I have nothing to add.

People are waiting for updates, we need to give them at least something :)

PR #96, opened 2022-11-10
Bunch of fixes and improvements

kurotatsu77:
Added replacement for standard dart traps, turning them into black goo traps, added UD_TrapManifest_ME_Script. NPCs are supported. 

Made UnforgivingSkyrim.esp patch for easier changes merge, to address GitHub inability to show changes inside esp. Suggest all contributors to create their own delta esps and commit them.

Fixed Abadon Cursed StraitJacket script item keyword from armbinder to straitjacket. 
Corrected name of Purified Goo magic effect

Changed Abadon Chastity Bra and Belt to looks shiny like cursed items from DCUR

Expanded FormID lists to include more hard bondage items

Wrote some massive changes into manifest random restraint procedure to make it use DD framework leveled lists and added harnesses and corsets. As result filters were changed massively, it was fixed to work again, but bit coding is changed. Abadon weapon and plug scripts were fixed to use proper filter values.

Changed devices in Abadon simple outfit - swapped iron boots for nicer looking black leather ones. 
Changed final punishing Abadon outfit, added more devices to be worn to ensure proper punishment.

Corrected some spelling and grammatical errors in messages.


IHateMyKite:
Nice work, will test it once I have some free time. The idea with .esp patches for github is also great. I think I will add Code of conduct so there is some unification between all the changes. For now, I would only ask you to rename UnforgivingSkyrim.esp so it also contains the number of pull request (so something like this -> PR96_UnforgivingSkyrim.esp) so it will be easier for me to merge it in the future. Once I merge it, I will just remove the patch.

kurotatsu77:
Yes, no problem.
Done.

kurotatsu77:
I think I marked this as ready for review, @IHateMyKite

Scripts/Source/UD_PurifiedGoo_AME.psc
```diff
@@ -6,9 +6,9 @@ UDCustomDeviceMain Property UDCDmain auto
 Event OnEffectStart(Actor akTarget, Actor akCaster)
     UDCDmain.DisableActor(akTarget)
     int loc_filter = 0xffffffff
-    if Utility.randomInt(0,1) ;50 % chance for suit
-        loc_filter = Math.LogicalAnd(loc_filter,0xFFFFFDFF)
-    endif
+;    if Utility.randomInt(0,1) ;50 % chance for suit
```
IHateMyKite, line 9: I think that previous implementation was better, as it allowed the player also see the belt/bra sometimes. Without this, the catsuit will always be equipped.

Scripts/Source/UD_RandomRestraintManager.psc
```diff
@@ -80,32 +80,43 @@ bool _mutex = false
...
-    UD_CheckKeywords.addForm(libs.zad_DeviousSuit)            ;iIndex = 9
-    UD_CheckKeywords.addForm(libs.zad_DeviousPlugVaginal)    ;iIndex = 10
-    UD_CheckKeywords.addForm(libs.zad_DeviousPlugAnal)        ;iIndex = 11
-    
-    UD_CheckKeywords.addForm(libs.zad_DeviousPiercingsVaginal)    ;iIndex = 12
-    UD_CheckKeywords.addForm(libs.zad_DeviousPiercingsNipple)    ;iIndex = 13
-    UD_CheckKeywords.addForm(libs.zad_DeviousGloves)            ;iIndex = 14
-    UD_CheckKeywords.addForm(libs.zad_deviousHood)                 ;iIndex = 15
-    
+
```
IHateMyKite, line 94: Rearranging the keywords will break previously created saves for older UD version. My current policy is to make new version UD backward compatible. If you still want to implement it like this, you will have to add and hook Update function, so when the game is loaded, the array is wiped and filled with rearranged values. In that case, also, please update the documentation which is at top of the script

Scripts/Source/UD_AbadonWeapon_AME.psc
```diff
@@ -18,7 +18,7 @@ Function OnEffectStart(Actor akTarget, Actor akCaster)
         if UDmain.TraceAllowed()
             UDmain.Log(GetActorName(akTarget) + " hit by abadon weapon. Locking device")
         endif
-        UDmain.UDRRM.LockRandomRestrain(loc_actor,false,0x00000080) ;lock hand restrain
+        UDmain.UDRRM.LockRandomRestrain(loc_actor,false,0x00000010) ;lock hand restraint
```
IHateMyKite, line 21: Either update the UDRRM or undo the changes (see comment in UDRRM)

Scripts/Source/UD_AbadonPlug_RenderScript.psc
```diff
@@ -188,18 +188,18 @@ Function randomEquipHandRestrain()
...
         else
-            UDCDmain.Print(getDeviceName() + " locks restrain on " + getWearerName())
+            UDCDmain.Print(getDeviceName() + " locks restraint on " + getWearerName())
         endif
     endif
 EndFunction
 
 bool Function equipRandomRestrain()
-    Armor device = UDmain.UDRRM.getRandomSuitableRestrain(getWearer(),0xFBFB)
+    Armor device = UDmain.UDRRM.getRandomSuitableRestrain(getWearer(),0xFFFCFFF3) ; exclude belts and plugs
```
IHateMyKite, line 197: Either update the UDRRM or undo the changes (see comment in UDRRM)

Scripts/Source/UD_TrapManifest_ME_Script.psc
```diff
@@ -0,0 +1,99 @@
...
+
+    if !akTarget.wornhaskeyword(libs.zad_deviousheavybondage)
+        UDCDmain.DisableActor(akTarget)
+
+		bool loc_haveSuit = akTarget.wornhaskeyword(libs.zad_deviousSuit)
+		Formlist loc_formlist = none
+	    Armor device = none
+        Int loc_arousal = UDCDmain.UDOM.getArousal(akTarget)
+
+;        if Utility.randomInt(1,99) < Round(UDCDMain.UD_BlackGooRareDeviceChance*fRange(loc_arousal/50,1.0,2.0))
```
IHateMyKite, line 22: Remove long comment block

Scripts/Source/UD_TrapManifest_ME_Script.psc
```diff
@@ -0,0 +1,99 @@
...
+        
+		if random > 90
+		    loc_formlist = UDCDmain.UDmain.UDRRM.UD_AbadonDeviceList_HeavyBondageHard
+        elseif random > 60 
+			loc_formlist = UDCDmain.UDmain.UDRRM.UD_AbadonDeviceList_HeavyBondage
+		else
+		    loc_formlist = UDCDmain.UDmain.UDRRM.UD_AbadonDeviceList_HeavyBondageWeak
+        endif
+
+        if loc_haveSuit
```
IHateMyKite, line 74: Rearrange the lines code, so the script is more readable

IHateMyKite (changes requested):
Please update the PR96_UnforgivingSkyrim.esp . The .esp is using the Cursed Loot textures. This means that the Cursed Loot is now UD master, which should not be the case. Either use different texture or undo the changes, so the Cursed Loot is no master.

Scripts/Source/UD_PurifiedGoo_AME.psc
kurotatsu77: Actually, this random is redundant, thanks to the keywords formlist rearranged, thus ensuring that HeavyBondage it applied first, and suit is after, only if that keyword available. All random calls are already inside RandomManifest.

Scripts/Source/UD_RandomRestraintManager.psc
kurotatsu77: Well, it has to be re-filled anyways because of added 2 keywords for corsets and harnesses.

I'm not sure how to add that Update function, any pointers for me?

Yep, will update the documentation so it will state clearly what is what. Makes total sense. I've left it in this state to be able to see "before" and "after" and assess changes.

Scripts/Source/UD_AbadonWeapon_AME.psc
kurotatsu77: Of course.

Scripts/Source/UD_AbadonPlug_RenderScript.psc
kurotatsu77: Yep.

Scripts/Source/UD_TrapManifest_ME_Script.psc
kurotatsu77: Ah, my blasted habit of avoiding removing old code and preserve it in comments...
Will clean it up.

IHateMyKite:
Also, one more point, please don't use tabs. They work badly on github. Use spaces instead. If you use Notepad++, you can just edit settings, so tabs are replaces by spaces (in case of UD by 4 spaces). You can also use Notepad++ to replace currently places tabs by replacing all \t characters with 4 spaces "    ".

IHateMyKite:
It is never a good idea to use tabs for programming, as different text editors will use different spacing for tabs. Because of that the text which allign fine for you will not allign good for other people (like me in this case because of how github handles tabs). By using spaces, you can never go wrong as the text will always be aligned correctly, no matter what text editor you use.

But in case of writing texts in Word, it is perfect habit as it saves you time.

Scripts/Source/UD_PurifiedGoo_AME.psc
kurotatsu77: > Please update the PR96_UnforgivingSkyrim.esp . The .esp is using the Cursed Loot textures. This means that the Cursed Loot is now UD master, which should not be the case. Either use different texture or undo the changes, so the Cursed Loot is no master.

Will copying TXTS from Cursed Loot, still pointing to the Cursed Loot files and removed master will be OK?
Or we definitely need to make copies of those files into UD assets?

kurotatsu77: OK, esp has no CursedLott as master anymore, made copies of TXTS and ARMA records, still using files from DCL. Is this OK?
Reverted change in the PurifiedGoo script, filter is set properly.

Scripts/Source/UD_RandomRestraintManager.psc
kurotatsu77: I think I figured out Update() function, added hook into main UD script, added necessary functions in UDRRM, tested in old save - it worked.

Cleaned up documentation, it's all written now as it should.

Scripts/Source/UD_AbadonPlug_RenderScript.psc
kurotatsu77: Done.

Scripts/Source/UD_AbadonWeapon_AME.psc
kurotatsu77: Done.
Tested, working.

Scripts/Source/UD_TrapManifest_ME_Script.psc
kurotatsu77: Cleaned it up, should be good now.

kurotatsu77:
Went through all changed scripts, replaced tabs with spaces.

Scripts/Source/UD_RandomRestraintManager.psc
```diff
@@ -275,7 +285,7 @@ bool Function LockAnyRandomRestrain(Actor akActor,int iNumber = 1,bool bForce =
     while iNumber
         iNumber -= 1
         loc_res = loc_res || LockRandomRestrain(akActor)
-		Utility.wait(3)	 ; small delay to ensure of no equips running into one another
+        Utility.wait(1)     ; small delay to ensure of no equips running into one another
```
IHateMyKite, line 288: This should be not needed, as lock function is mutexed. Adding this will make locking the devices slower with not additional safety added.

Scripts/Source/UD_RandomRestraintManager.psc
```diff
@@ -342,7 +352,7 @@ int Function LockAllSuitableRestrains(Actor akActor,Bool force = false,Int iPref
             endif
         endif
         loc_i += 1
-		Utility.wait(3)	 ; small delay to ensure of no equips running into one another
+        Utility.wait(3)     ; small delay to ensure of no equips running into one another
```
IHateMyKite, line 355: Same as previous comment

IHateMyKite (changes requested):
The issue with textures can't be solved by pointing at the textures. That will solve the maser issue, but the retextured devices will lack textures if user have no DCUR installed. I think that better way of doing this will be to hook the texture set to previous used texture (or other one, but it must be in base DD). Then, the DCUR patch can be updated so it also edits the texture sets. This way, if user will have DCUS patch installed, the DCUR textures will be used, Otherwise, previous textures will be used.

kurotatsu77:
Done, as bonus I've made MCM page for user-friendly device filter selection.

Scripts/Source/UD_RandomRestraintManager.psc
```diff
@@ -51,68 +53,85 @@ Function StartMutex()
...
     endif
     ready = True
 EndEvent
 
+Function Update()
+    FillOutCheckKeywords()
+    if UDmain.TraceAllowed()    
+        UDCDmain.Log("Refilled Keywords formlist")
+    endif
+    UD_RandomDevice_GlobalFilter = 0xFFFFFFFF
```
IHateMyKite, line 105: This line should not exist. This will wipe all MCM settings which you added on every game reload.

Scripts/Source/UD_RandomRestraintManager.psc
```diff
@@ -162,51 +181,87 @@ Armor Function getRandomDeviceByKeyword(Actor akActor,Keyword kwKeyword)
...
+    elseif kwKeyword == libs.zad_DeviousCorset && !akActor.wornhaskeyword(libs.zad_DeviousHarness)
+        LL = zadDL.zad_dev_corsets
+    elseif kwKeyword == libs.zad_DeviousHarness && !akActor.wornhaskeyword(libs.zad_DeviousCorset)
+        LL = zadDL.zad_dev_harnesses
+    else
+        return none
+    endif
+
+    if LL
+        int tries = 10                        ; 10 attempts
```
IHateMyKite, line 252: Reduce this number to ~2. If it failed first time, there is big chance it will fail again on next attempt. So forcing this check 10 times can freeze player for few minutes before all conflicts are resolved. You should also add some console trace, so users can see which devices are conflicting. In this case, I think the best option is UDmain.Warning or just GWarning

IHateMyKite (changes requested):
Both of following issues are in esp, so you can skip them and I will fix them when I will be merging them. Or you can do it yourself, I will keep it on you. Once you fix the 2 script related issues in UD_RandomRestraintManager, I will merge the PR.

[PR96_UnforgivingSkyrim.esp](https://github.com/IHateMyKite/UnforgivingDevices/pull/96/files#diff-b18eaf4593540e767c05dcc2c36647b884c74c2240244eb0fe91e8f3426e0f72)
The formlists should not include Abadon Cursed Straitjacket. Device should be Abadon quest exclusive. Making the device 100% sentient is also crazy, as it will make escaping it almost impossible. 

[PR96_UnforgivingDevices - DCL Patch Patched.esp](https://github.com/IHateMyKite/UnforgivingDevices/pull/96/files#diff-9f1425d667655d79c1639c8d2383500bca82236b896f9b763f9d55f57c963db5)
We didn't understand each other, why I meant was that this esp should only change the texture set, not the device. If it change device, it will mean that every time the base device is changed, the patch device will also need to be updated. Which is pain, and its easy to forget to do it, which results in unnecessary bugs.

Scripts/Source/UD_RandomRestraintManager.psc
kurotatsu77: Oh... *facepalms*

Scripts/Source/UD_RandomRestraintManager.psc
kurotatsu77: Actually, my testing shown that this reroll works really fast!
It's the most useful when pulling various harnesses, some can have collars, some can have chastity.
So it tries several times to pull variety of device which is not in conflict with others.

Maybe leave it as it is for testing, and if some really nasty hold-ups arise, then reduce?

kurotatsu77:
Try it, current Cursed SJ 100% sentient can be escaped easier than Rare Blue Straitjacket.
Also in description there's line about it always activating plugs on escape attempt...
If you feel it's quest-only thing - sure thing, set it as you feel it should be.

Scripts/Source/UD_RandomRestraintManager.psc
IHateMyKite: I have tested it and it took very long time for all devices to be locked. But it might have been caused by something else. Will have to investigate further later. For now, I will ignore this point. After merge, I will have to do some more benchmark testing.

IHateMyKite (approved):

IHateMyKite:
Nice work. Closing and merging.

PR #100, opened 2022-12-01
Experimental AI Control

IHateMyKite:
Experimental AI control. Added it as PR, so it doesn't affect main branch for now.


kurotatsu77 (changes requested):
First things first - I absolutely LOVE this feature!
Tested some, here's some things to fix:
1. Better move check of helper being tired to the first reply, i.e. right after being asked for help and not at the moment when they try to actually do something. Quality of Life thing.
2. Details for NPC are not being shown, something broke?
3. I think Abadon Plug motivation should be lowered, so NPC would try to escape all other items first, should be easier the escape Abadon Plug then.
4. This needs a lot of grammar love, but it's no problem, I can go through it and add PR later.
5. What about "Tighten" action? I failed to see any result despite receiving message about tightening restraint...

IHateMyKite:
>     2. Details for NPC are not being shown, something broke?

What did you use to show the details ? NPC menu or device ?

kurotatsu77:
> > ```
> > 2. Details for NPC are not being shown, something broke?
> > ```
> 
> What did you use to show the details ? NPC menu or device ?

NPC menu, called via ; , didn't worked for other NPCs and for follower. Also I believe player's details didn't worked either.

Devices details work fine.

IHateMyKite:
That is strange, I can't replicate the bug. Does opening the details show something or does it shows nothing ? It should show menu like this 
snap

IHateMyKite:
Try to recompile all scripts. Alternatively there is mod with all compiled updated scripts 
[UD - PR100.zip](https://github.com/IHateMyKite/UnforgivingDevices/files/10279623/UD.-.PR100.zip)

kurotatsu77:
> That is strange, I can't replicate the bug. Does opening the details show something or does it shows nothing ? It should show menu like this snap

OK, redid the testing, results are following:
1. Player details are shown, called by menu through button . 
2. NPC via "Details" menu, called by button ; is not working - nothing shown. If I call Details on unregistered NPC, NPC gets registered, but no Details menu shown.
3. If I go through inventory and call up inventory item's menu and select Details / Wearer's details - I get them as they should be.

To sum it up - only Details page for NPCs are not shown when called through menu by hotkey ; on the NPC.
I've double-checked it that I've compiled all the scripts and whatnot.

Finally I've replaced UnforgivingDevices.esp with one from this branch and it started working. This means the main esp has some changes which differs from the parent and weren't been put into the delta esp?

kurotatsu77 (approved):

IHateMyKite:
They should be the same (well, once they get merged to latest, this branch is now behind the main esp). I converted the esps to xml, and only changes are in the grammatical errors you fixed, nothing else, so the esp should be not cause of issue. Check that the patch is in load order after the main mod.

PR #101, opened 2022-12-03
Adapting iWidget

iiw2012:
I adapted wide-screen fix to iWidget canvas and tested it on several screen resolutions.
There is test placement of the widgets and reper points in InitWidgets function.

Also changed 'old' widgets to remove duplicates and make them cleaner. Namely, I replaced UDCustomHeavyBondageWidget1.psc with UD_WidgetBase. Added new property (PositionYOffset) to control widget vertical position relative to group.


IHateMyKite:
Thank you very much for your sacrifice. Tried it and all X are on correct places. Once this is merged, I can at least add more widgets for different statuses. For now I'm thinking about adding new one for device condition (so the durability bar will no longer change color). Maybe another one for Abadon plugs strength when the quest is active. Possibilities are truly infinite.

iiw2012:
Always glad to help!
Looks like I have nothing to add here. The technical part is more or less in order, what remains is the implementation of your ideas, and that I leave to you :)

IHateMyKite (approved):
Ok

IHateMyKite:
I have tried to implement the changes, but there is some issue with  meter widgets. Basically, the test ones work as intended, but the ones used by minigames are behaving strange. Everytime the Toggle is called for iWidget, it will make widget to be positioned to the right. It also looks more streched. I think that is somehow caused by SetPercentage function. I will have to test this furthere, but this looks like bug from iWidget itself.

IHateMyKite (changes requested):
On hold

iiw2012:
I tried using setMeterPercent on test indicators and it works. 

```
            If True
                ; Meters on bottom position
                ; Widget on the left above magica meter
                Int m = iWidget.loadMeter()
                iWidget.setSize(m, HUDMeterHeight as Int, HUDMeterWidth as Int)
                iWidget.setPos(m, CalculateGroupXPos(0), (CalculateGroupYPos(0) - HUDMeterHeightRef * 1.5) As Int)
                iWidget.setVisible(m, 1)
                iWidget.setMeterPercent(m, 25)
                
                ; Widget at the center above health meter
                m = iWidget.loadMeter()
                iWidget.setSize(m, HUDMeterHeight as Int, HUDMeterWidth as Int)
                iWidget.setPos(m, CalculateGroupXPos(1), (CalculateGroupYPos(0) - HUDMeterHeightRef * 1.5) As Int)
                iWidget.setVisible(m, 1)
                iWidget.setMeterPercent(m, 50)

                ; widget on the right above HUD stamina meter
                m = iWidget.loadMeter()
                iWidget.setSize(m, HUDMeterHeight as Int, HUDMeterWidth as Int)
                iWidget.setPos(m, CalculateGroupXPos(2), (CalculateGroupYPos(0) - HUDMeterHeightRef * 1.5) As Int)
                iWidget.setVisible(m, 1)
                iWidget.setMeterPercent(m, 75)

                ; widget on the right above the last one
                m = iWidget.loadMeter()
                iWidget.setSize(m, HUDMeterHeight as Int, HUDMeterWidth as Int)
                iWidget.setPos(m, CalculateGroupXPos(2), (CalculateGroupYPos(0) - HUDMeterHeightRef * 3.0) As Int)
                iWidget.setVisible(m, 1)
                iWidget.setMeterPercent(m, 100)
            EndIf
```

ScreenShot8_2

Could it be drawShapeLine? Because I don't really understand what it does and left it on its place.

Could you please commit your code that didn't work? Even if it's not functional.

IHateMyKite:
In my case, pushed code result in this 
Widget

iiw2012:
I don't know why but when I moved setMeterPercent after UpdateGroupPositions it worked. Apparently, this is another "feature" of the Skyrim interface.

After experimenting, I found out that the setPos and setSize functions conflict with setMeterPercent. If there was a call of setMeterPercent on a widget, then any further attempts to change its size or position (even if it doesn't really change anything) corrupt its dimensions.

I modified the UpdateGroupPositions function so that new widget instances are created each time it is called. This is not a perfect solution, but I managed to change positions from MCM without distortion.

IHateMyKite (approved):

IHateMyKite:
Did some adjustments. Looks like everything works fine now. Thank you for your work.

PR #106, opened 2022-12-20
Grammar fixes

kurotatsu77:
I went through all scripts and esp and did my best to straighten up grammar whenever I could. So far result looks quite good to me, but of course suggestions and additions from English native speakers are highly welcome.


IHateMyKite (approved):
Nice work

PR #107, opened 2022-12-22
A bit of grammar love

kurotatsu77:
Some small fixes, also please note when merging esp - UD_DefaultCustomPlug_interactionMsgNPC [MESG:18150D9E]
has changes from "DO nothing" to "Do nothing", and looks like it's not detected as changed as it's only one letter register change.


IHateMyKite:
Will have to fix the long menu later, will create some submenus for certain thinks. Something like button for debug, button for NPC register (will have register, unregister and persist/unpersist buttons), etc...

IHateMyKite:
Will have to fix the long menu later, will create some submenus for certain things. Something like button for debug, button for NPC register (will have register, unregister and persist/unpersist buttons), etc...

IHateMyKite (approved):
Nice work

PR #112, opened 2023-01-15
all done

kurotatsu77:
Made minimum values of skill advance and skill effect to 0. Added For Him esp detect, checkbox to allow devices to be used on men at DD Patch tab. Changed ValidForUD function to respect this setting. Added checks into all appropriate magic effects to make sure they won't work on men unless allowed. Couple of grammar fixes while at it.

### DIFF
--- a/Scripts/Source/UD_MCM_script.psc
+++ b/Scripts/Source/UD_MCM_script.psc
@@ -39,8 +39,8 @@ int little_finisher_chance_S
 int min_orgasm_little_finisher_S 
 int max_orgasm_little_finisher_S 
 
-int  dmg_heal_T
-int  dmg_magica_T
+int dmg_heal_T
+int dmg_magica_T
 int dmg_stamina_T
 
 int AbadonForceSuitOnEquip_T
@@ -659,6 +659,8 @@ int UD_StartThirdpersonAnimation_Switch_T
 int UD_DAR_T
 Int UD_OutfitRemove_T
 Int UD_CheckAllKw_T
+Int UD_AllowMenBondage_T
+
 Event resetDDPatchPage()
     UpdateLockMenuFlag()
     setCursorFillMode(LEFT_TO_RIGHT)
@@ -681,8 +683,8 @@ Event resetDDPatchPage()
     AddHeaderOption("Device setting")
     addEmptyOption()
     UD_OutfitRemove_T = addToggleOption("Outfit removal", UDCDmain.UD_OutfitRemove)
+    UD_AllowMenBondage_T = addToggleOption("Allow DD devices on men", UDmain.AllowMenBondage,FlagSwitch(UDmain.ForHimInstalled))
 EndEvent
-
 
 UD_WidgetControl Property UDWC Hidden
     UD_WidgetControl Function Get()
@@ -863,6 +865,10 @@ Function resetOtherPage()
     
     AddTextOption("SlaveTats installed: ",InstallSwitch(UDmain.SlaveTatsInstalled),FlagSwitch(UDmain.SlaveTatsInstalled))
     addEmptyOption()
+
+    AddTextOption("Devious Devices For Him: ",InstallSwitch(UDmain.ForHimInstalled),FlagSwitch(UDmain.ForHimInstalled))
+    addEmptyOption()
+
 EndFunction
 
 String Function InstallSwitch(Bool abSwitch)
@@ -1082,6 +1088,9 @@ Function OptionDDPatch(int option)
     elseif option == UD_OutfitRemove_T
         UDCDMain.UD_OutfitRemove = !UDCDMain.UD_OutfitRemove
         SetToggleOptionValue(UD_OutfitRemove_T, UDCDMain.UD_OutfitRemove)
+    elseif option == UD_AllowMenBondage_T
+        UDmain.AllowMenBondage = !UDmain.AllowMenBondage
+        SetToggleOptionValue(UD_AllowMenBondage_T, UDmain.AllowMenBondage)
     elseif option == UD_CheckAllKw_T
         UDmain.UD_CheckAllKw = !UDmain.UD_CheckAllKw
         SetToggleOptionValue(UD_CheckAllKw_T, UDMain.UD_CheckAllKw)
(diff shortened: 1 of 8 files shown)


Scripts/Source/UD_MCM_script.psc
```diff
@@ -2218,6 +2227,8 @@ Function DDPatchPageDefault(int option)
         SetInfoText("Prevent NPC outfit from being removed when hand restraint is locked on. Removing outfit can by default cause compatibility issue with NPC overhaul mods. This will obviously prevent NPC from being naked until player undresses them\nDefault: True")
     elseif option == UD_OrgasmAnimation_M
         SetInfoText("List of orgasm animations.\nNormal = Normal orgasm animations by  DD\nExtended = Orgasm animations + horny animations\nDefault: Normal")
+    elseif option == UD_AllowMenBondage_T
```
IHateMyKite, line 2230: There is no need to add this part, as the whole Default part is still WIP. Because of that, I kept the previous info texts as template for future use (so there is no need to write all the ifs again). Will have to finish it myself or create request

Scripts/Source/UnforgivingDevicesMain.psc
```diff
@@ -132,6 +132,9 @@ bool Property SlaveTatsInstalled        = false auto
 bool Property OrdinatorInstalled        = false auto
 bool Property ZadExpressionSystemInstalled = false auto
 Bool Property DeviousStrikeInstalled    = False auto
+Bool Property ForHimInstalled           = False auto
+
+Bool Property AllowMenBondage           = False auto
```
IHateMyKite, line 137: IMHO, I think that this should be by default **True**. If someone have For Him installed, it should already be taken as that they want to use the bondage on males. The more rare case is  if someone have For Him installed and don't want bondage on males (in which case, they should use the MCM setting, not the other way around)

Scripts/Source/UnforgivingDevicesMain.psc
```diff
@@ -524,6 +537,9 @@ bool Function ActorIsValidForUD(Actor akActor)
     if loc_race.IsChildRace()    ;check that actor is not child
         return false
     endif
+    if (loc_actorbase.GetSex() == 0 && !AllowMenBondage)
```
IHateMyKite, line 540: This part should also check if For Him is installed. Only if both For Him is installed and it is enabled, it should proceed. Also, please move the loc_actorbase.GetSex() == 0 on the end of if, so it is last condition. This way, it will not check actor sex, if the For Him is not installed (aka, optimization).

IHateMyKite (changes requested):
As mentioned, please take a look at the commented parts.

Scripts/Source/UD_MCM_script.psc
kurotatsu77: Well, this doesn't bad, right? And will be useful later to make sure this is not forgotten, right?

Scripts/Source/UnforgivingDevicesMain.psc
kurotatsu77: Current men support even with For Him installed is quite so-so. This is why I opted to set it to Disable.
Also set to False is much easier to implement and more robust, as it handles properly all enabling and disabling esps.

For example, I have For Him enabled, but I prefer to play with this disabled, much better results. And UD's for men support is essentially non-existent.

Scripts/Source/UnforgivingDevicesMain.psc
kurotatsu77: Actually, I've put 
else
        ForHimInstalled = false
        AllowMenBondage = false
    endif

in the esp detection function, so it ensures it will be forced to be disabled, making this quite robust. So one less condition to check (as it's sure to be false if esp is not there), so maybe not good idea?

Swapping places - no problem, will do.

Scripts/Source/UD_MCM_script.psc
IHateMyKite: You are right. Sorry about that.

Scripts/Source/UnforgivingDevicesMain.psc
IHateMyKite: But this isn't only about Abadon quest, but about devices in general. ActorIsValidForUD is also used when locking devices, so no devices can be locked on actor if it's false. 
And you are missing one important point, if you had For Him installed when using DD, you knew that it meant that the devices will work for men. So why should it now be reversed ? People will start to flood the support thread asking why For Him is suddenly not working, and you would have to explain to them that they have to actually enable it. So please, change it, so it is by default true. As I said before, the number of people using For Him, and not wanting bondage on man is much smaller than the number of people using For Him, so they can use devices on males.

Scripts/Source/UnforgivingDevicesMain.psc
IHateMyKite: That it true, but I think it should still be an other way around, as it makes code more clean to read. The ActorIsValidForUD is quite important function in UD, so it can be assumed that other people might trey to take a look at it in the source code. And if they see the condition as it is now, they will not realize that it is actually linked to For Him .esp. Also, checking bool is one of the fastest operations you can do in Papyrus, so it will not have that much of the impact.

So, I think the condition should look like this
if (!ForHimInstalled && !AllowMenBondage && loc_actorbase.GetSex() == 0)
    return false
endif

[This ](https://github.com/IHateMyKite/UnforgivingDevices/blob/8858c7014db0fcc0ba1283c9e25404503737ff75/Scripts/Source/UnforgivingDevicesMain.psc#L417) part will then be irrelevant

Scripts/Source/UnforgivingDevicesMain.psc
kurotatsu77: Got your point, will change it.

Scripts/Source/UnforgivingDevicesMain.psc
kurotatsu77: Yep, this will look like it.
Guess we'll place 3 conditions check instead of two, if you insist, no problem.

IHateMyKite (approved):
Nice work

PR #113, opened 2023-01-21
Interface widgets 02

iiw2012:
I did not plan to post this PR in the current version, but it so happened that I found and fixed several errors for the current release, so I propose to consider my additions in their entirety :) Unnecessary or unstable parts can be turned off at any time.

Added an overlay using iWidget:
- animated icons of equipped vibrators.
- animated effect icons: active orgasm, fatigue after orgasm, fatigue after struggling.
- display of UD, DD notifications in a specified zone on the screen.

The overlay can be configured via MCM:
- change the general position of the icons: left, right and around the crosshair.
- change the position of the notifications vertically.
- enable or disable different parts of the overlay.

Icons are animated:
- the color reflects the strength of the effect (strength of vibration, number of fatigue stacks).
- blinking signals the current active effect (active vibrator, lasting orgasm).

All done through the added API in UD_WidgetControl:
- Notification_Push(String asText)
- StatusEffect_SetVisible(String asName, Bool abVisible = True)
- StatusEffect_SetMagnitude(String asName, Int aiMagnitude)
- StatusEffect_AdjustMagnitude(String asName, Int aiAdjustValue, Bool abControlVisibility = True)
- StatusEffect_SetBlink(String asName, Bool abBlink = True)

Also improved or changed:
- legacy widgets now flashes properly;
- update layout on property changes;
- switching UD_WidgetControl script states when configuring MCM;
- CurrentVibStrength implemented like property.

Once again I was convinced that the architecture of device scripts is built very competently. Adding these effects was elementary. It was much more difficult to deal with iWidget.

It is recomended to start from save without active iWidget. Because different versions of UD_WidgetControl can overlap (???), resulting in undefined behavior.

<details>
<summary>NSFW Examples</summary>

Near the center

overlay_exhaustion

overlay_orgasm

On the left

overlay_test
</details>

I'm ready for feedback. :)


iiw2012:
I'm trying to make transition between different widget engines smooth, but for now the best practice is to save/load game after switching.

IHateMyKite:
This is incredible. It is certainly a step in the right direction, as it solves a bunch of existing UI issues (too many messages, etc…). But I have question regarding the messages, I see the DD notifications show up too. How does this work ? Did you patch the DD scripts, or is there some change to native notification function ?

Also, IMA opinion, the orgasm widget is too much complex, and it took me a second to realize what it depicts. There is no need to show genitals in it (as it is quite complex drawing in this case). From far a way, it looked to me like some kind of alien egg xD. Maybe some simpler widget, with like heart or something, might be enough. But that might not be good depiction of orgasm...

iiw2012:
> Did you patch the DD scripts, or is there some change to native notification function ?

It is all in zadlibs_UDPatch.psc. It turns out that it is enough to override a few functions to reroute (almost) all DD notifications.

> Also, IMA opinion, the orgasm widget is too much complex, and it took me a second to realize what it depicts.

I had doubts about that image too. I'll try to find something else.

iiw2012:
Tried to change icons for effects. Since I am not an artist, I have to use what can be found on the net.

<details>
<summary>Variant 1 (ahegao style)</summary>

ScreenShot30

</details>

<details>
<summary>Variant 2 (also changed icon for exhaustion to be in the same style)</summary>

ScreenShot32

</details>

Variant 2 is in the last commit.

kurotatsu77:
Variant 2 looks better.

IHateMyKite:
I also think that the second variant orgasm icon is better. Its very explicit, so it should be immediately clear what it means. On the other hand, I think that the exhaustion icon from 1st variant was better. Maybe it could be possible to just make it toggle in MCM, so users can choose which icon to use ?

iiw2012:
> Maybe it could be possible to just make it toggle in MCM, so users can choose which icon to use ?

At the moment I'm using filenames as effect identifiers in API functions. To implement the ability to change icon for the effect, I will have to add the translation of the identifier to the file name. It's not very difficult. But maybe we'll be able to achieve the ideal right away. :)

<details>
<summary>Variant 3 (changed icon for the exhaustion effect, thicker lines for better looking in bright scenes)</summary>

ScreenShot34

</details>

kurotatsu77:
Kneeling person is better exhaustion depiction IMHO, so I still vote for Variant 2.
If adding ability to select variant for user is not too much of hassle - it probably better to be done, because everyone's taste differs and providing ability to choose is nice.

IHateMyKite:
I also thing that the sitting one is better for exhaustion, but problem is that it still doesn't very much match the exhaustion feeling. Maybe adding the 3 wiggly lines like in the case of the first variant might be enough. Like this 
pic

iiw2012:
Changed the exhaustion effect icons as suggested above. And added options in the MCM to select icons variants.

Along the way, added the ability to register custom statuses, which will be displayed either with devices or with effects.

I plan to improve the graphic part of the overlay:
- add a translucent background under the icons for a better look in different scenes.
- add an outline for the letters of the text.

IHateMyKite:
Found one issue. Loading the save will always set the icon variant to the 1 variant. Most likely caused by [this](https://github.com/iiw2012/UnforgivingDevices/blob/489cd82773e982cd35723db1d5a59b304e57e262/Scripts/Source/UD_WidgetControl.psc#L269-L275) part

IHateMyKite:
Also found that there are many errors in papyrus log related to the arrays
[Papyrus.0.log](https://github.com/IHateMyKite/UnforgivingDevices/files/10517802/Papyrus.0.log)

iiw2012:
Spamming errors in the log is probably due to saving with a previous version of the PR. I tried to fix it, but it's best to start with a clean (no PR) save. I'll see how best to solve this problem of loading saves with older versions in the future.

For saves from version 2.0 everything should work.

IHateMyKite:
I have started a new game (but I also tried to load the existing save and issue was the same). Will try the updated version to see if the issue is still there.

iiw2012:
I like it, much better than before! 
I saved all textures as dds with DXT5 compression to preserve the transparency channel. I could do it for you since I have all tools ready. :)

iiw2012:
A bunch of parallel arrays for icon parameters is already too complex to maintain. Therefore, I think to implement this part as an array of the quest aliases. There will be a finite number of icon slots, but the data will be in structures. 
I believe it will be easier to add or remove fields during development while maintaining save compatibility.

What do you think?

IHateMyKite:
> I like it, much better than before! I saved all textures as dds with DXT5 compression to preserve the transparency channel. I could do it for you since I have all tools ready. :)

Thanks but that is not necessary, I already have some experience with editing dds files, because I was recoloring DD textures. I'm using GIMP for that, and it works as intended. I have also generated the mipmaps, even through I dunno if they are needed in this case (but files have same size as before, so it should have same attributes).

> A bunch of parallel arrays for icon parameters is already too complex to maintain. Therefore, I think to implement this part as an array of the quest aliases. There will be a finite number of icon slots, but the data will be in structures. I believe it will be easier to add or remove fields during development while maintaining save compatibility.
> 
> What do you think?

I think that in this case, it is truly the only way forward, because the amount of arrays is already too big. You can just add aliases to quest, and then you can add some get function for them. The aliases can be named as WidgetSlots or something like that. You can take a look at the NPC Manager which use similar system. The only difference is that in this case, the aliases will not be filled with actors, but empty.

iiw2012:
At the moment I have implemented almost everything that I wanted. There is one thing left that makes me doubt: the transfer of interaction with the meter widgets to similar API calls. 

Something like that:

```
Function Meter_Register(String asName)
Function Meter_SetVisible(String asName, Bool abVisible, Bool abUpdateVisPos = True)
Function Meter_SetPercent(String asName, Int afValue, Bool abForce = false)
Function Meter_SetColor(String asName, Int aiColor, Int aiColor2 = 0, Int aiFlashColor = 0xFFFFFF)
Function Meter_Flash(String asName)
```

In addition to making the API more consistent, it will be possible to implement the ability to register and use additional widgets from external modules. Just like now it is done with icons.

What do you think?

IHateMyKite:
I think it's ok, it is explicit and understandable, so others should have no issue with that.

IHateMyKite:
Scripts can't be compiled, because there is missing script UD_WidgetMeter_RefAlias. Did you perhaps forget to put it under source control?

IHateMyKite:
Found a strange bug. After commit 9fd956c, the esp somehow gets broken. It no longer have UD as master, and as so it is adding its own quest instead of overwriting them. Looking at it, the formIDs somehow get changed to the order of the patch. Don't know if this is possible to be fixed with xEdit. Best way might be to use esp before the change, and reimplement back the changes

iiw2012:
> - Updated PR113_InterfaceWidgets.esp to use forms 43 instead of forms 44 (44 is only for SE, most likely caused by use of SE CK or SSEdit)

I'm using SSE CK all the time. I didn't think it would create any problems, since LE CK, Tes4Edit and the game open plugins without visible issues.

I won't be editing UnforgivingDevices.esp with SSE tools, and for plugins I'll either use `SetFormVersion(e, 43);` or will also switch to LE tools if there will be errors.

IHateMyKite:
Are there any more changes planned ? I have tested it and didn't find any issues. Only in convention might be that the durability and condition widget is swapped (red one for durability should be down, while green conditional one should be above orgasm meter).

iiw2012:
I have corrected the order of the widgets. Changes should appear after starting a new game.

I had an idea to add small icons next to the meters to identify the "resource" they are measuring. If you don't mind, I'll try to implement it in the coming days.

Scripts/Source/UD_CustomVibratorBase_RenderScript.psc
```diff
@@ -327,7 +327,7 @@ bool Function canVibrate()
 EndFunction
 
 bool Function isVibrating()
-    return _currentVibRemainingDuration != 0 && CurrentVibStrength
+    return _currentVibRemainingDuration > 0 && CurrentVibStrength > 0
```
IHateMyKite, line 330: This implementation will not work for vibrators which have infinite duration. Infinite vibration is done by simply specifying the duration as negative. The duration then gets reduces on every update, but it will never reach 0. Is there any specific reason why was it changed ?

iiw2012: > Is there any specific reason why was it changed ?

I've encountered a bug where the vibrator turns off but the corresponding icon doesn't update (remaining colored and blinking).

So I randomly corrected all the places that seemed doubtful to me. And I didn’t come across that bug anymore, but this is probably just a coincidence. :)

Thanks for correcting me, I'll check everything again.

IHateMyKite:
Can confirm there are no more issues with vibrators. Starting new game show widgets in correct order and the icons next to them shows also correctly. Will do some more testing and if no issue is found, I will close the PR (unless you have some more changes in mind).

iiw2012:
I didn't have any problems either. The latest icon changes are not as noticeable as I thought. I barely pay attention to them during mini-game. :) So it's time to close this PR.

PR #114, opened 2023-01-21
Device lock rework

IHateMyKite:
Total rework of device locks. All devices will now have set number of locks (depending on device type). 

- The number will not change. Every lock can have different difficulty and accessibility. 
- All locks can also have lockpick shields. Lock can't be unlocked with lockpicking until all shields are removed. Basically, if device have 5 shields, it means that lock needs to be lockpicked 6 times for it to be removed.

Whole idea is that for example, the catsuit can have 5 locks, 1 on back, 2 on arms and 2 on legs. And every lock can have different accessibility (back being hardest to reach). Unlocking locks now also provide stronger boost to struggle minigames (up to 300%, might add MCM setting for it)

### DIFF
--- a/Scripts/Source/UD_CustomDevice_EquipScript.psc
+++ b/Scripts/Source/UD_CustomDevice_EquipScript.psc
@@ -193,10 +193,12 @@ Event OnContainerChanged(ObjectReference akNewContainer, ObjectReference akOldCo
 EndEvent
 
 Function openWHMenu(Actor akTarget,Actor akSource)
-    if UDmain.TraceAllowed()    
-        UDCDmain.Log(" NPC menu opened for " + deviceInventory.getName() + " (" + akTarget.getActorBase().getName() + ") by " + akSource.getActorBase().getName(),1)
+    if UDmain.IsEnabled()
+        if UDmain.TraceAllowed()
+            UDCDmain.Log(" NPC menu opened for " + deviceInventory.getName() + " (" + akTarget.getActorBase().getName() + ") by " + akSource.getActorBase().getName(),1)
+        endif
+        UDCDMain.OpenHelpDeviceMenu(getUDScript(akTarget),akSource,UDCDmain.ActorIsFollower(akTarget))
     endif
-    UDCDMain.OpenHelpDeviceMenu(getUDScript(akTarget),akSource,UDCDmain.ActorIsFollower(akTarget))
 EndFunction
 
 UD_CustomDevice_RenderScript Function getUDScript(Actor akActor)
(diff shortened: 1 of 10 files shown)


kurotatsu77:
Interesting idea. Maybe make shields not only lock-based but timer-based too?
It can be lock shield, timer shield or both.
I think I saw request to implement timer lock-shields like in the base DD at LL.

IHateMyKite:
That is valid question. I forgot about it. But adding it should be possible. All locks are done as bitcoded ints, so there is a limit in how many things I can put there. If I add the time unlock, it will be in hours. I will reserve 7 bits for that, so the limit will be 127 hours. Hopefully that is enough. Alternatively, it can be in 2 hours (1 = 2 hours). This way, time lock can't be less than 2 hours, but it can be up to 254 hours. Or maybe even more hours. Let me know what you think.

kurotatsu77:
It's coded as hours in DD, I think we should use the same, i.e. 1=1 hour.
127 hours as limit sounds about right to me.

IHateMyKite:
Most things are now done.
Things changed:
- All devices now use different lock system
- Existing device will be incompatible, so unlock them first before testing
- Patcher should correctly add the locks to patched devices
- All Abadon devices should also correctly use new lock system
- Added Lock Details to details tab
- Added TimeLock to locks. Some locks can be timelocked, which prevent to manipulate them
- Created simple tool for creating new locks on https://ihatemykite.github.io/LockBitcoder.html

kurotatsu77:
I think for better review this should be updated to match main branch, please.

IHateMyKite:
Working on it right now. There are already few things I forget to update (NPC AI so it works with new locks), so I need to edit it anyway

IHateMyKite:
Done, but I didn't test the Ai lockpick minigame. Testing the AI is in general quite a pain in the ass, so it might be better to just push the publish beta and see what users find.

kurotatsu77 (commented):
1. Rope devices can have locks and locks details. Probably DD patcher needs to be changed so when device is rope - it should have no locks.
2. Unlock option in the Lock actions is shown when no locks possible to unlock, observed when locks are UNLOCKED/JAMMED/JAMMED/JAMMED for example. Probably some checks need to be improved.
3. Current lock system is a bit tiring because of need to perform actions on each lock separately. Proposed QoL improvement - make each action to go at all available locks. I.e. when trying to unlock - it will try to unlock all possible to unlock locks one after another (not jammed, not timelocked etc). Same with locks repair and lockpicking. Same should be done to AI struggling too, because they go to cooldown on each lock opened, which makes their escape go really slow.

kurotatsu77:
Few more findings:
1. When trying to perform any action on lock (unlock, lockpick, repair) and select BACK instead of lock, struggle minigame gets started instead of simply closing menu.
2. Default number of lock shields is set to 6, in MCM menu it's set to 1. Probably need to bring MCM menu to speed?

IHateMyKite:
1,2,4,5= fixed
3 = I don't thing that is possible with current design. Ever locky have different accessibility and difficulty. Both of these values are used when degerminating crit chance or repair progress added per crit. Because of that, new minigame needs to be started for every lock. You can imagine that every lock is to some extent a very small device on the main device.

kurotatsu77 (approved):
It's great work, long needed. 
I do suspect #3 will be raised by users, but it does makes sense to go at locks separately.

Also I have small request for you, when you merge esps, please find Message FormID xx150D9E with EditorID UD_DefaultCustomPlug_interactionMsgNPC. In it's Menu Button #8 text is "DO nothing". Please replace it with "Do nothing". I've submitted this change couple of times, but as it's only case change, it doesn't gets detected as change and never got into the main esp.

PR #116, opened 2023-01-25
Added lock durability

IHateMyKite:
Simple PR. Added key durability and optional PO3 Papyrus extender integration (https://www.nexusmods.com/skyrimspecialedition/mods/22854)

- Key durability = Every time key is used to unlock lock, its durability will reduce. Once it reaches 0, it will get removed. This value can be edited in MCM Custom Device tab (Main section)
- PO3 Papyrus Extender integration = Currently to disable player detection, ConsoleUtil is used. Issues is that I have feeling that it is causing game freeze when in lockpick minigame. Out of all freezes, 95% of them were invoked when lockpick minigame ended (or when one lock was lockpicked, basically, when lockpick container was closed with some time delay). I have feeling that the ConsoleUtil detection toggle is source of the issue. If the issue will still continue even after using PO3, I will the detection part, and debug it further.
Note that now the sources for the PO3PE will be required to compile the UD source scripts


kurotatsu77 (commented):
I've tested it on LE, I wonder if it's good enough. Maybe ask someone on SE/AE to look at this too, maybe with these optional mods added?

On my LE it worked good, no major problems spotted, added feature working.

One small thing noticed, could be missed during test of previous PR - when selecting device action (unlock, struggle, etc) from open inventory window, it stays open. Action starts after user closes the window. I suspect it's not desired behavior and maybe needs to be addressed.

IHateMyKite:
The menu thing is a long time issue. Basically, The UIExtensions is not working correctly while in menu. So there are 2 possible fixes. Either I do it like you said, and make all menu to be closed if UIExtensions menu is opened. OR, I can try to create new script based on UIExtensions script, but with the menu compatibility. But, it is possible that the 2. option might not be possible, its more of experiment. Will try to do the 2. one first ,as having inventory closed every time you try to see details about device is painfull.

kurotatsu77 (approved):
I guess my review on LE was enough, let's merge this.

IHateMyKite:
I started working on the UIExtension script override, and it looks promising. The List menu is working as intended when in other menu. Only issue is that it is not backward compatibility, so new game is required or resaver needs to be used to delete old objects which cause issues. I will try to find a way to make it backward compatible and then push it.

IHateMyKite:
In case you wan't to make existing save compatible, you need to do following 

- Load the existing save. Wait for UD to update. Don't open any menus while in this state, as it can break the same! After UD is updated, save and exit the game immidiatly.
- Open resaver and delete this form
hint

kurotatsu77 (approved):
Worked like a charm!
This save cleaning instruction worth to be to put inside the patch notes for the users.

PR #125, opened 2023-02-09
Performance update

IHateMyKite:
- Updated performance of minigame by changing some values from bitcode to direct acess. AV are now check in paralel to main loop. Only works for player, NPC will have AV checked directly on main loop
- Updated how prebuild expressions are stored


kurotatsu77 (approved):
At very least it's not slower, maybe a bit faster than before.
Still, can mess up things really bad by assigning special key to mousewheel down and run cut game, then unlatch clicker on the wheel and ROLL it. :)
Maybe add some kind of limiter of how many clicks per second can be made to failsafe from it?

IHateMyKite:
Added one quick fix, and the check you mentioned. Didn't test it yet tho, so maybe it will not work

IHateMyKite:
Tested it, and it works. Merging

PR #128, opened 2023-03-09
NPC Static slots

IHateMyKite:
Implementation of NPC static slots. This should allow to add new NPC slots which can be used by quest. Still WIP


kurotatsu77:
I think I get the idea of what you're trying to do here, but I can't understand the way for me to test this thing out...
I thought custom NPC won't take normal slots? Nodaba in the test esp takes one when bound.

How can I verify and see this thing working?

IHateMyKite:
Sorry, might have give more information.
What I basically did was that now other mods can add they own NPC Slots and fill them with npcs. But in comparison with base UD, these slots are static. This means they can't be changed by player. Only through scripts or esp. They can't also be filled with scanned npcs. My base idea is that moder can register their npc in 2 ways.

1.  By directly filling the npc slot alias (as in this case). This can be prefered way if the npc should be always registered. I have used Nodaba more like example, not that there is really reason to register here.
2. By using framework functions. In that case the Slots needs to be first fetched by function GetStaticSlots from NPC manager (one linked in UDmain).

So for example, if you want to do the quest in which some npc should wear devices, and you also want them to be also updated, and use AI, you can do it this way. More practical example might be what we talked about before in ideas. If there would be for example the quest with Nodaba where she would get locked in devices, you can register here right away. This way,  she can actually escape the devices before even players arrive to her, because of AI. All other effects will be also used, like device activations, orgasm management etc..

So basically, the main reason why I did it is to allow moders to register npcs without worrying about the npc slots. If the UD slots were used, then it can be filled by player, in which case it can break the quest because  all slots might be used. It will also make quest smother, because users don't have to worry about registering npcs.

Also, regarding the test. If you meet the Nodaba, she should be automatically registered. Issue is that it will require new game because of changed factions. Or you can add her to the factions (UD_RegisteredNPC_Faction and UD_StaticNPC_Faction). If then open the NPC Menu on her, you should not see Register/Unregister button. Otherwise she should work exactly like registerted NPC.

kurotatsu77:
Ah, that's why she was auto-added to normal slots, because I didn't ran it with new game? Got it, will try to get a good look tomorrow.

IHateMyKite:
Most likely. If you add here to the factions using cmd (addtofaction or setfaction rank command), it should work.

kurotatsu77:
Loaded save before goo research quest introduced into load order, worked fine. Nodaba will not appear in the detected npc list when bound, however she successfully escaped her restraints despite player spending time in another cell. 
Only one minor bug spotted - while waiting in Arcadia's shop, I was able to hear struggle moans from Nodaba in the Bannered Mare Inn, probably something to address?

IHateMyKite:
Will have to give it a look. For now, I did some quick update, but I don't think that will fix the issue. Will have to find the source of the issue first.

kurotatsu77:
Yep, tested it, still can hear struggle moans in another cell. Minus that it works perfectly.

IHateMyKite:
Did you leave the cell when she was already struggling ? Or did it start on its own. Are you also possibly close to the house ?

kurotatsu77:
I waited for her struggle to end, she was in Bannered Mare. After that I left to Drunken Huntsman, so essentially 2 cells away, should be far enough. And I waited for a while, checking console, moans appeared when she was struggling.

IHateMyKite:
Will investigate further. Have to do some debugging

IHateMyKite:
Hm tested it, and the panting can't be heard, even if I go at the other side of skyrim. But I have one crazy theory which might explain it. All interiors are basically stored in one big cell. So I think that it is possible that Arcadia shop and Bannered mare is close to each other in interior space. This way, the Nodabas 3d can be loaded even when player can't effectively reach her. Can you check if you can hear it when you go to outside (tamriel) cell. Going outside of whiterun should be enough.

kurotatsu77:
Turned out your theory is right on spot! I can hear her faint moans inside of some internal cell of Whiterun but not other places.
And can't hear her while standing directly outside of Bannered Mare doors! 
I like this feature, it will be very useful.

kurotatsu77 (approved):
I like this feature, it will be very useful.

Faint moaning can be heard only in adjacent (in the same worldspace) internal cells, so it's perfectly fine, maybe note this as known behavior or simply ignore it.

PR #134, opened 2023-03-29
Minigame duration adjustments

IHateMyKite:
Created PR of existing sb, so it gets more exposure and I don't forget about it


kurotatsu77 (approved):
I think everything is in place here, great work!

PR #135, opened 2023-04-02
updated the MCM script file and added MCM language files

YanJunSunYJS:
updated the MCM script file and added MCM language files

Fix typos, unify text

need new game or clean save to display text correctly!

### DIFF
--- a/Scripts/Source/UD_MCM_script.psc
+++ b/Scripts/Source/UD_MCM_script.psc
@@ -252,7 +252,7 @@ EndEvent
 String _lastPage
 Event OnPageReset(string page)
     if !UDmain.IsEnabled()
-        AddHeaderOption("Unforgiving devices is updating...")
+        AddHeaderOption("Unforgiving devices is updating or disabled...")
         return
     endif
     
--- a/Scripts/Source/UnforgivingDevicesMain.psc
+++ b/Scripts/Source/UnforgivingDevicesMain.psc
@@ -464,27 +464,42 @@ Bool Function IsEnabled()
     return !_Disabled && !_Updating && ready
 EndFunction
 
+;Previous sattes of scripts, so they are returned to correct state
+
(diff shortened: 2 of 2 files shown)


IHateMyKite:
Thank you for your contribution. Tested it works as intended. Only issue I found is that changing some string in translation file have no effect in game

- $General
- $NPCs Config
- $Other

My guess is that other mod is using the same string and so invalidate the changes made to the UD file. Might be better idea to name all string in same manner as it is done later with UD_ prefix

YanJunSunYJS:
> Thank you for your contribution. Tested it works as intended. Only issue I found is that changing some string in translation file have no effect in game
> 
> * $General
> * $NPCs Config
> * $Other
> 
> My guess is that other mod is using the same string and so invalidate the changes made to the UD file. Might be better idea to name all string in same manner as it is done later with UD_ prefix

Tested on my end no problem.
If it can be explained, it means that other mods you installed contain these strings, and these corresponding strings have been localized, for example, we say $General here, if other mod also involve it, and translate it into other languages Text, then $General will definitely not be displayed as General, there is a sequence problem here. As long as the interface does not display $General

IHateMyKite:
Found out some of the same string are in skyui_se_english.txt. I think that best solution would be to add prefix, so it will be impossible for the strings to get replaced by other translation file
[skyui_se_english.txt](https://github.com/IHateMyKite/UnforgivingDevices/files/11131749/skyui_se_english.txt)

YanJunSunYJS:
> Found out some of the same string are in skyui_se_english.txt. I think that best solution would be to add prefix, so it will be impossible for the strings to get replaced by other translation file [skyui_se_english.txt](https://github.com/IHateMyKite/UnforgivingDevices/files/11131749/skyui_se_english.txt)

OK, i will change them :)

PR #138, opened 2023-04-02
Merge latest changes, modify optimized language files

YanJunSunYJS:


IHateMyKite:
Everything works except for $UD_NPCSCONFIG. I'm assuming this is because UD_NPCsPageName is used. 

https://github.com/IHateMyKite/UnforgivingDevices/blob/018bee7d8df7f6b43b62bc3e1273e74ca139ebab/Scripts/Source/UD_MCM_script.psc#L132

Can you try to remove the variable and replate it with the string ?

YanJunSunYJS:
> Everything works except for $UD_NPCSCONFIG. I'm assuming this is because UD_NPCsPageName is used.
> 
> https://github.com/IHateMyKite/UnforgivingDevices/blob/018bee7d8df7f6b43b62bc3e1273e74ca139ebab/Scripts/Source/UD_MCM_script.psc#L132
> 
> Can you try to remove the variable and replace all of it calls with the $UD_NPCSCONFIG string ?

I just tested the MCM text display in the game and there is no problem, all texts are displayed correctly, especially the one you mentioned

IHateMyKite:
Strangely, it now works. Most likely caused by me not making new game.

IHateMyKite:
Nice work

YanJunSunYJS:
> Strangely, it now works. Most likely caused by me not making new game.

If the translation reference text definition has changed significantly, you must use a new game or a clean document.
for example
If it is defined as **_$Test_** in the script file, it is defined as **_$Test Test_** in the language file
The new definition is **_$UD_Test_**, and its disciple is defined as _**$UD_Test Test**_ in the language file
If you don't start a new game or clear your game saves, the text on the UI doesn't display correctly

PR #141, opened 2023-04-08
Fixed several typos in mcm language files

YanJunSunYJS:


IHateMyKite:
Nice work, but please, do not rename the scripts. This is only thing that can never be done the version is release as it means that it will be not backward compatible. It will also break all other mods which use this mod. Please, undo the change, so it it named again UD_ControlablePlug_RenderScript

Also, you don't have to recycle the PR every time there are some things to change. That makes it harder for me to see if relevant things were fixed.

PR #143, opened 2023-04-08
Some fixes

YanJunSunYJS:
In the future, I will only be responsible for the maintenance of language files, and I don’t care about the others, it involves too much


IHateMyKite (approved):
Nice work

PR #151, opened 2023-05-01
Fixes for initialization process in UDWC

iiw2012:
When loading a script in non-default state, default methods are called anyway for some reason (and even OnEndState("iWidgetInstalled") is called). Because of this, the initialization of iWW could occur twice.

I removed unnecessary calls from the default state functions. And added code to the OnEndState event handler, which allow switching between engines without save/load.

I tried the changes in the game, and haven't found any problems yet.

### DIFF
--- a/Scripts/Source/UD_WidgetControl.psc
+++ b/Scripts/Source/UD_WidgetControl.psc
@@ -419,8 +419,8 @@ Event OnUpdate()
             return ;fatal error, do not use the module
         endif
         Ready = True
-        OnUIReload(abGameLoad = True)
-        InitWidgetsRequest(abGameLoad = True, abMeters = True, abIcons = True, abText = True)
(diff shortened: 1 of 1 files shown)


IHateMyKite:
Tested it and mostly works without issue. Only issue I found is that the config setting is ignored on new game. I have auto load ON and Use iWW ON. Before the changes, every time I started the game, it always immediately started using iWW. Now, it for some reason always use legacy widget system. I had to switch turn it off (which did nothing, because there was nothing to do), and then re-enable it so it can start using iWW. So either there will need to be change to how its initiated, or the config setting can be ignored (which would mean that users will always have to toggle iWW ON manualy)

iiw2012:
I'll take a look.

IHateMyKite (approved):
Nice work. No more issues found.

PR #157, opened 2023-05-19
Experimental:  Native function - Minigame drain

IHateMyKite:
Started learning how to make SKSE plugins. This is basically the first experiment in that regard. It replaces the way how actor values are reduced and checked while in minigame. Instead of doing it in papyrus, it is done with native function. This makes minigame much more smooth, as the stats are reduced at the same rate as in vanilla game.

Some important notes

- This is done only on SE. It uses Common Libs NG and Adress Library, and so should be compatible between AE ans SE. But I can't test it as I dont have AE
- It will certainly not work with LE, and I'm not planning on converting it to it. Whole community for dll plugins on LE is basically dead, so I dont have any resources to use
- In this case I made it so it use the native function always. In case it will work as intended, I might add some switch so it can be turned off when used on LE

I will be happy for any feedback. As I said, this is the first thing I have done, so crashes are not out of question (my game crashed like 50 times before I get it right)


Ethryan:
**_AE_** = **_SE_** the only thing that’s different is that _**SE**_ is running on an older version of the game.

IHateMyKite:
That not so simple, as the game binaries different. That is why there are suddenly so much different mods for different versions. Many things were moved and so its very easy for dll to break by just existing as it will be accessing/calling different data then it was intended.

iiw2012:
I've tested it on Skyrim AE and it works fine. I believe the difference in binaries is compensated with corresponding SKSE Address Library.

IHateMyKite:
I think that most of the work is done. Here is final list of changes

- Device minigame and orgasm minigame will use native function for reducing actor stats. This makes minigame much smoother
- Added checks that should prevent from native library to be used if user have not correct version of SKSE or if they have no library at all
- Library will be not part of main repo, but part of fomod repo (as having it present on LE build will create issues, so LE users should have option to not have it installed)
- All sources for native plugin will be in new repo [https://github.com/IHateMyKite/UnforgivingDevicesNative](https://github.com/IHateMyKite/UnforgivingDevicesNative)
- Use of native library can be toggled in MCM (but only if correct SKSE version is installed and if plugin is present)
- (hopefully) Compatible SKSE versions are 2.0.20 (SE) and 2.2.3 

Some additional changes that are not related to native plugin
- NPC manager should now update much faster

Please, everyone who gets the time, give it a fly and add the review/feedback. If it will work without issue, I will merge it to main branch

iiw2012:
I've run tests on AE 1.6.353 and everything works fine. I had to change the version check condition by adding 0.2.1 for that.

It is probably safe to weaken the version check. I didn't see any changes in the latest changelog except fixes and adding compatibilities.

```papyrus
    ;check if correct skse version is installed
    if SKSE.GetVersion() == 2
        ;is version 2.0.20 (SE)
        if SKSE.GetVersionMinor() == 0 && SKSE.GetVersionBeta() == 20
            return true
        endif
        ;is version 2.1.x (old AR)
        if SKSE.GetVersionMinor() == 1
            return true
        endif
        ;is version 2.2.x (AR)
        if SKSE.GetVersionMinor() == 2
            return true
        endif
    endif
```

IHateMyKite:
> ```papyrus-skyrim
> SKSE.GetVersion() == 2
> ```

Good point. I will just make it to check only major and minor version. No need to check beta version.

PR #161, opened 2023-05-27
Native function - Native widget meter handling

IHateMyKite:
In this PR, I added native functions which replaces how widget meters are handled. They are no longer updated using UI script, but instead using the custom native functions which directly edit the value. It works in following way

- Either SkyUi default meter or IWW meter have to be registered using AddMeterEntryX
- Once that is done, meters will internally change their value based on set rate
- Current value of meter can be read with GetMeterValueX
- This way, no calculation/update of widget have to be done in papyrus script. Instead, only rate is changed. This way the widget update will be unhooked from papyrus speed/lag. 
- Issue is that the actual new value have to be read from the said widget. Because of that, it was needed to do some big changes to papyrus.
- As far as I tested it, it should work for both default meters and iWantWidgets. 
- Yet again, this feature is tied to Use native variable, and because of that it can be disabled mid game. Tested it and it can even be disabled without reloading the game.

Some additional fixes that I did while adding this feature
- Yet again fixes some properties pointing at old UD_Libs (most notably, Black Goo effect script)
- Changing Native support and IWantWidget support in MCM will now force mod to update. (for safety reasons)

For now I think this is all. Depending on feedback, I might do some addition changes. Otherwise this PR is ready for review.
@iiw2012 As this PR mostly tough the UD_WIdgetControl script and its use in the mod, I would love to hear your feedback opinion.


iiw2012:
Wow! You've done a great job here!

Have you considered changing the UDWC API by adding new functions?
```papyrus
Function SetMeterRate(string asMeterName, float afNewRate)
Function SetMeterMult(string asMeterName, float afNewMult)
Function SetMeterValue(string asMeterName, float afNewValue)
```
Then all the conditions checks and private function calls could be hidden there. These are just thoughts, because I haven't delved deeply into all the changes yet.

IHateMyKite:
> Wow! You've done a great job here!
> 
> Have you considered to change the UDWC API by adding new functions?
> 
> ```papyrus-skyrim
> Function SetMeterRate(string asMeterName, float afNewRate)
> Function SetMeterMult(string asMeterName, float afNewMult)
> Function SetMeterValue(string asMeterName, float afNewValue)
> ```
> 
> Then all the conditions checks and private function calls could be hidden there.
> 
> These are just thoughts, because I haven't delved deeply into all the changes yet.

Sounds like good idea. Will do some changes so its more easy to use and add functions to Widget control

IHateMyKite:
I think that the work is mostly done for now. Somebody, please test it and let me know if it works without issues.

@iiw2012 I was thinking about moving some animation related methods to native plugin, so the system works faster. Issue is I'm not sure which part is slowest. Any hint? Also, the whole JSON thing can also be done in C++ and it will be much faster. Other things, like checking keywords and math, too. So it should be possible to move everything to C++, but the issue is if some part requires some variables from papyrus.

iiw2012:
I used to check the performance of various UDAM functions, and the slowest and most frequently used one turned out to be GetConstraintsInt. I even had to cache the result of this function. Other calls are within the norm with some margin. Therefore, I do not see the need to transfer json functions to the SKSE plugin.

But keyword checks take a very long time. If you replace vanilla keyword functions with faster ones, this will be a big improvement, and not only in the animation subsystem.

I suggest checking the following functions:

- `Int GetActorConstraintsInt(Actor)` - calculates the bit mask of constraints for the actor.
- `Keyword GetHeavyBondageKeyword(Actor)` - returns the keyword for heavy bondage on the actor.

These functions use keywords from the DD, so it probably will be tricky to implement them in the plugin. I haven't seen any complaints about UD performance yet, so you can leave all as it is if the changes require a lot of work.

iiw2012 (approved):

IHateMyKite:
> I used to check the performance of various UDAM functions, and the slowest and most frequently used one turned out to be GetConstraintsInt. I even had to cache the result of this function. Other calls are within the norm with some margin. Therefore, I do not see the need to transfer json functions to the SKSE plugin.
> 
> But keyword checks take a very long time. If you replace vanilla keyword functions with faster ones, this will be a big improvement, and not only in the animation subsystem.
> 
> I suggest checking the following functions:
> 
>     * `Int GetActorConstraintsInt(Actor)` - calculates the bit mask of constraints for the actor.
> 
>     * `Keyword GetHeavyBondageKeyword(Actor)` - returns the keyword for heavy bondage on the actor.
> 
> 
> These functions use keywords from the DD, so it probably will be tricky to implement them in the plugin. I haven't seen any complaints about UD performance yet, so you can leave all as it is if the changes require a lot of work.

In my case, when playing the animation with helper, it takes quite a long time to start. Solo animation works without issue. Is there any main difference between them ? Also, keywords can also be checked in skse plugin and its much faster. I will close this PR and give it a look in the new PR. There will be only small changes on papyrus side so it should be easier to implement.

iiw2012:
> In my case, when playing the animation with helper, it takes quite a long time to start. Solo animation works without issue. Is there any main difference between them ?

In terms of JSON reading there is not much difference. Delays may appear, as it is necessary to disable and align the second actor. But again there is not much differences from the player.

I have a built-in benchmark on the animations page in MCM. It measures the time taken to get a list of suitable animations from JSON - this is the most time-consuming stage when it comes to reading files. You could measure time with the following settings:

ScreenShot53

It should give the maximum amount of suitable animations.

ScreenShot54

This time is the upper limit, because in the menu list building is 1.5 times slower for some reason.

PR #173, opened 2023-08-13
Fixed couple of issues related to sorting device slots

blu3mania:
Fixed the problem that sortSlots exits prematurely in the rare case when consecutive slots become empty when executing removeUnusedDevices. In theory the same problem could also happen when unregisterDevice is called and multiple copies of that device exist in consecutive slots.

Also fixed the problem that sortSlots gets called many times unnecessarily when removeCopies is executed, because unregisterDevice always invokes sortSlots even if nothing is actually removed.


IHateMyKite (approved):
No issue

IHateMyKite:
Great job! Thanks for your contribution.

PR #178, opened 2023-10-07
Version 3.0 Implementation

IHateMyKite:
- Every commit to branch dev-22 will only work on SE. As mentioned before in discussion, from this point on, UD will be SE only. Version 2.1 will be last LE supported version (+ quick fix once it gets merged)
- Reworked modifiers from ground to make working with them easier. Adding new modifiers should be also more simple.
- Added new native functions and added them to papyrus code
- Removed all native switches (for LE compatibility). This might in some places greatly increase performance (mainly in device operations)
- Branch is names 22 but version will be 30. Will rename once this is merged


IHateMyKite:
Experimental release 1: [Unforgiving Devices - Experimental 2.2 NG.zip](https://github.com/IHateMyKite/UnforgivingDevices/files/13031981/Unforgiving.Devices.-.Experimental.2.2.NG.zip)
Will work best with DDNG

kurotatsu77:
I think the best thing would be calling SE only UD v3.0. This way you'll get all possible 2.* version numbers for bugfix releases.
Less confusion and thus less questions from users.

IHateMyKite:
Might have added it sooner, but most discussion about this version (including releases of experimental build) is done on new DDNG discord https://discord.com/invite/VCkBRbJ6sb

Ethryan:
Add the link to the discord to the main github readme page.

IHateMyKite:
> Add the link to the discord to the main github readme page.

Nice point. Will do

Scripts/Source/UD_Modifier_Demanding.psc
```diff
@@ -0,0 +1,41 @@
...
+        return true
+    endif
+
+    Int loc_gold = CalculateGold(aiDataStr,akModDevice.UD_Level,false)
+    Int loc_WearerGold = akModDevice.GetWearer().GetItemCount(UDLibs.Gold)
+    
+    Bool loc_cond = loc_WearerGold > loc_gold
+    
+    if !loc_cond && IsPlayer(akModDevice.GetWearer())
+        UDmain.Info("You dont have enaugh gold to pay the device!")
```
kurotatsu77, line 18: enough

IHateMyKite: Thanks, will fix

Scripts/Source/UDCustomDeviceMain.psc
```diff
@@ -3504,9 +3408,128 @@ Int[] Function DistributeLockShields(Int aiLockNum, Int aiLockShieldNum)
...
+    Parameters:
+
+        akDevice        - Device to check
+
+    Returns:
+
+        True if device is from steel or iron
+/;
+bool Function isSteel(UD_CustomDevice_RenderScript akDevice) global
+    return (StringUtil.find(akDevice.deviceInventory.getName(),"Steel") != -1) || (StringUtil.find(akDevice.deviceInventory.getName(),"Iron") != -1)
```
kurotatsu77, line 3507: Also there's "primitive" and "locking" iron plugs.
Many steel chastity items have "Lustrous" in their names.
Also there are colored chastity items, like Black, Silver, Golden, White, Red. How to handle these - I can't think up...
And the worst part of using the item names is problems with translations.
Interestingly enough, in the DD Integration there are material keywords added, like zad_Material_Metal, which would fit here right at home, but they are not applied to any of the restraints. Unfinished feature?
Also logical addition would be adding color keywords too, but I guess it's not implemented yet.

IHateMyKite: Nice point. First I will rework it so it is native. I will also add new config setting so user can select which words are needed in name for it to be of specific material. I can also additional make it to check for zad_Material keyword. Dunno why it was not used, but might be great point for NG discussion

PR #179, opened 2023-11-03
Dismount for ImprovedCamera

AndrejAndb:
Tested with different CameraState. 
Has some problems with free camera (looks in rare cases, and not only on a horse), but poor knowledge of the mod does not allow to test in detail


IHateMyKite:
Nice job

PR #183, opened 2024-02-15
Version 3.0 Implementation

IHateMyKite:
Renamed branches, so previous PR got basically removed. It can be seen here https://github.com/IHateMyKite/UnforgivingDevices/pull/178


Ethryan (commented):
The UD_Config file seems to be missing from this pull request when i compare it to the previous one.

The number of edited lines is also different when you compare the two pull requests.

IHateMyKite:
> The UD_Config file seems to be missing from this pull request when i compare it to the previous one.
> 
> The number of edited lines is also different when you compare the two pull requests.

That is because there were some additional PR in between, which were then merged in. While this tries to merge directly to main branch. Either way, if you look at UD_Config in the last commit, you will see it is same as the one from previous PR

Scripts/Source/UDItemManager.psc
```diff
@@ -13,31 +14,35 @@ Event OnInit()
...
+    if !akTarget.wornhaskeyword(libs.zad_DeviousPiercingsNipple) && UDmain.UDRRM.IsDeviceFiltered(1)
         libs.LockDevice(akTarget,UDmain.UDlibs.AbadonPiercingNipple)
     endif
 EndFunction
 
 Function lockAbadonHelperPlugs(Actor akTarget)
-    if !akTarget.wornhaskeyword(libs.zad_deviousPlugAnal) && !AbadonScript.UseAnalVariant
-        if Utility.randomInt(0,1)
+    if !akTarget.wornhaskeyword(libs.zad_deviousPlugAnal) && !AbadonScript.UseAnalVariant && UDmain.UDRRM.IsDeviceFiltered(3)
+        if RandomInt(0,1)
```
kurotatsu77, line 32: Maybe add Abadon Quest completion here, thus allowing random plugs without risk of them evolving into Abadon Plug?
Something like if RandomInt(0,1) && AbadonQuest.IsComplete()?

Scripts/Source/UD_AbadonQuest_script.psc
```diff
@@ -115,15 +122,93 @@ String[] Property UD_AbadonSuitNames
...
+            if _NextHandRestrain > 4
+                _NextHandRestrain = 0 ;loop
+            endif
+            if IsPlayer(akTarget)
+                UDmain.ShowMessageBox("Black goo covers your body and tie your hands while changing shape to bondage restraint!")
+            endif
+        endif
+        
+        ;chance to lock plugs and belt. Only works if player first finished abadon quest (as some plugs can evolve in to abadon plug)
+        if IsCompleted() && RandomInt(1,99) < Round(15.0*fRange(loc_arousal/25.0,1.0,3.5))
```
kurotatsu77, line 189: As stated in previous comment, this quest completion check can be moved into the lock plugs function. Thus even without quest it still can allow to get other plugs.

Scripts/Source/OrgasmSystem.psc
```diff
@@ -0,0 +1,103 @@
+Scriptname OrgasmSystem hidden
```
kurotatsu77, line 1: Not something like UD_OrgasmSystem?

UnforgivingDevices.esp
kurotatsu77, line 1: Will double the Discord report so won't be forgotten:
1. Please consider adding boxbinder outfit to act as evolve target for non-outfit one. I'll switch to using yours in US then.
2. Please consider making topless ebonite SJ to evolve into normal ebonite SJ, it would make a lot of sense.
3. Cursed Inflatable Plug needs balancing
4. Abadon Clitoral Piercing has MAH modifier instead of MAO.
5. Preventing to use same device from black goo but remaining random is great idea. Also please replace messagebox for console message after first times.

Will give it a whirl tomorrow ingame to double-check if I missed anything else...

kurotatsu77 (approved):
All in all, I believe it's quite good to merge, all notes I've posted can be addressed later if you wish so.

UnforgivingDevices.esp
kurotatsu77: Also spotted another thing: Abadon High Heels evolving into Abadon Iron Boots. This doesn't looks like logical way, I'd expect them to evolve into Abadon Ballet boots, same black and shiny,

Scripts/Source/OrgasmSystem.psc
IHateMyKite: The idea was to make the system as its own mod in the future, but it doesn't look like it will happen right now. Nevertheless, it can be imagined as small mod inside UD

Scripts/Source/UDItemManager.psc
IHateMyKite: Good point. Will do

kurotatsu77 (commented):
These changes work like a charm, really good!

Meanwhile found another minor thing: when helper helps to unlock restraints, after unlocking first lock, minigame still continues without selecting the next lock to unlock. If this was the last lock - it works as it should.

IHateMyKite:
> Meanwhile found another minor thing: when helper helps to unlock restraints, after unlocking first lock, minigame still continues without selecting the next lock to unlock. If this was the last lock - it works as it should.

Cant reproduce. Does this happens all the time, or only sometimes ?

kurotatsu77:
> > Meanwhile found another minor thing: when helper helps to unlock restraints, after unlocking first lock, minigame still continues without selecting the next lock to unlock. If this was the last lock - it works as it should.
> 
> Cant reproduce. Does this happens all the time, or only sometimes ?

Rechecked with the latest dll and all recompiled scripts.
a. When player helps follower to unlock - everything works as it should, locks are opened one after another, when one lock opened - dialogue appears to select the next as it should.
b. When follower help player to unlock their restraint, after unlocking first lock minigame continues, no dialogue to select the next lock appears, hitting crits does nothing. So to unlock lock you have to start another unlock game once per lock remaining, and when the last lock is opened - restraint removed and minigame ends properly.

This effect appears each time without fail when player receives unlocking help from follower (helper). When player unlocks themselves or NPC - everything always works perfectly.

kurotatsu77:
Please see https://github.com/IHateMyKite/UnforgivingDevices/issues/184

IHateMyKite:
> Please see #184

Will fix with new PR. Closing this now, so I can separate different issues/features in different PRs

PR #185, opened 2024-03-11
Fixed plug minigame not working correctly with helper

### LINKED ISSUE #184: PR#183 Bug on plug removal with helper
**Describe the bug**
Minigame to remove plug from player not working correctly when helper is used

**To Reproduce**
1. Wear a plug
2. Get help from helper to remove it
3. Observe bar not filling, thus impossible to get success.

**Expected behavior**
Normal plug removal minigame expected

Skyrim AE GOG 1.6.659 
UD from PR#183 and latest UDNative.dll

IHateMyKite:
- Fixes #184


kurotatsu77:
Plug minigame completely broken for me, both solo and with helper. Do I need to run new game to test this?

IHateMyKite:
Ehat exactly is broken ? New game is not required, i tested it on existing save without issue.

kurotatsu77:
> What exactly is broken ? New game is not required, i tested it on existing save without issue.

All device bars are moved to side, power bar not raising.

IHateMyKite:
Found out that I forget to switch branch while pushing changes to fomod repo, so the old dll was used instead of the new one. Should be now fixed. Test the NB once it finishes

kurotatsu77:
I'm getting crazy...
ScreenShot7

Nothing helps. I even went as far as compiling UDnative,dll, got it in size of 1.5 Mb instead of 4.8 from the zip file though. No matter what dll I use - the same result... Gotta sleep on it, I guess...

kurotatsu77 (approved):
Yep, works as it should now.

And the problem that I was having? SLUTS Resume 3.3... As soon as I reverted to 3.2 - all went normal again.

Sorry for wasting a lot of your time on this.
Still got to wonder, why my compiled UDnative.dll is only 1.5 Mb in size though?

IHateMyKite:
> Yep, works as it should now.
> 
> And the problem that I was having? SLUTS Resume 3.3... As soon as I reverted to 3.2 - all went normal again.
> 
> Sorry for wasting a lot of your time on this. Still got to wonder, why my compiled UDnative.dll is only 1.5 Mb in size though?

Does the SLUTS use native plugin ? If yes ,then that might explain few issues, as there might be some incompatibility with UD.

kurotatsu77:
> > Yep, works as it should now.
> > And the problem that I was having? SLUTS Resume 3.3... As soon as I reverted to 3.2 - all went normal again.
> > Sorry for wasting a lot of your time on this. Still got to wonder, why my compiled UDnative.dll is only 1.5 Mb in size though?
> 
> Does the SLUTS use native plugin ? If yes ,then that might explain few issues, as there might be some incompatibility with UD.

Nope, no plugin.

PR #186, opened 2024-03-15
Added SetPlayerFollower. Updated Help menu

IHateMyKite:
- Added SetPlayerFollower function which forces actor to follow player. This also forces actor to stop any other package (this uses alias). This was copied from ORomance, which was using this function to make npc follow player untill they start scene
- This is now used in Help menu. So if you want help from npc, or want to help npc, the npc should no longer just walk away, or sit on the closes chair, preventing minigame from starting


kurotatsu77:
Got UDnative.dll crash while testing this, was using PRNB only, no additions.

kurotatsu77 (approved):
All looks good, helper actually following player adds nice touch. Small stutter observed on initial "receive help" dialogue, but nothing major.

IHateMyKite:
> All looks good, helper actually following player adds nice touch. Small stutter observed on initial "receive help" dialogue, but nothing major.

This is intended. I added small wait time so npc have time to get up from furniture if they are using it. But I didnt really test if its needed. Will test it and possibly remove it

PR #190, opened 2024-04-11
Added more events

IHateMyKite:
- Added UD_Events, and some basic events created by UD
- Rebalanced cutting/plug minigames again

Please let me know if you have some ideas about additional events

### DIFF
--- a/Scripts/Source/OrgasmSystem.psc
+++ b/Scripts/Source/OrgasmSystem.psc
@@ -3,6 +3,7 @@
 Scriptname OrgasmSystem hidden
 
 ;/   About: OrgasmVariable
+     --- Code
      enum OrgasmVariable : uint8_t
      {
          vNone                       =  0,
@@ -28,18 +29,22 @@ Scriptname OrgasmSystem hidden
      
          vLast
      }
+     ---
 /;
 
 ;/   About: OrgasmUpdateType
+     --- Code
      enum OrgasmUpdateType : uint8_t
      {
         mSet                        = 0x1,
         mAdd                        = 0x2,
         mMultiply                   = 0x3,
      }
+     ---
 /;
 
 ;/  About: OrgasmMod
+    --- Code
     enum OrgasmMod : uint32_t
     {
          //default modes
@@ -56,9 +61,11 @@ Scriptname OrgasmSystem hidden
 
          //16-31 = Duration (seconds)
     }
+    ---
 /;
 
 ;/  About: EroZone
+    --- Code
     enum EroZone : uint32_t
     {
         eNone                   = 0x00000000,
@@ -73,6 +80,7 @@ Scriptname OrgasmSystem hidden
         eAnal2                  = 0x00000100,
         eDefault                = 0x00000200    //when you dont care about ero zones
     }
+    ---
 /;
 
 Bool    Function AddOrgasmChange(Actor akActor, String askey, Int aiMod,Int aiEroZones, Float afOrgasmRate = 0.0, Float afOrgasmRateMult = 0.0, Float afOrgasmForcing = 0.0, Float afOrgasmCapacity = 0.0, Float afOrgasmResistence = 0.0, Float afOrgasmResistenceMult = 0.0) global native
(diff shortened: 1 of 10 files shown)


kurotatsu77:
A thought - could be useful to get references of the items and actors, because getting base forms out of references is easy, but not the other way.
Events themselves look quite ok to me.

garryrot:
So, as you maybe know I'm implementing a mod that can control bluetooth devices with buttplug.io. Maybe I just haven't figured out how to do this, but I would like some body part specific events, so that I can do some better "immersion" with that.

Currently, OnVibrateEffectStart and OnVibrateEffectStop only gives you the info that something is vibrating, and the vibration strength, the default DD events do not make any distinction which body part specifically is affected.

The widgets, however, seem to know the following body parts:

image

You've seem to have already added these:

    IflatablePlug_Inflate,
    IflatablePlug_Deflate

Some UD-Specific events would be cool like:

    PlugVaginal_Start, PlugVaginal_Stop
    PluginVaginal_Inflate, PiercingVaginal_Deflate

    PlugAnal_Start, PlugAnal_Stop
    PluginAnal_Inflate, PiercingNipple_Deflate

    PiercingClit_Start, PiercingClit_Stop
    PiercingNipple_Start, PiercingNipple_Stop


StrArg could be info about the pattern type, if its on, off etc..
FloatArg could be info about the strength of the vibration

BTW: Yes, the library I use (bp.io) can control inflators, constrictors (do the opposite of inflators), oscillators, rotators, vibrators, strokers, so the more metadata those events can contain, the merrier.

IHateMyKite:
> A thought - could be useful to get references of the items and actors, because getting base forms out of references is easy, but not the other way. Events themselves look quite ok to me.

I think that it is not possible to send references with events (at least the not persist ones). But currently the events send both ID and RD, so it should be possible to use UD functions to get the reference. For example getDeviceByInventory or getDeviceByRender on UDCDmain




> So, I've created a mod that can control bluetooth devices with buttplug.io. Maybe I just haven't figured out how to do this, but I would like some body part specific events, so that I can do some better "immersion" with that.
> 
> Currently, OnVibrateEffectStart and OnVibrateEffectStop only gives you the info that something is vibrating, and the vibration strength, the default DD events do not make any distinction which body part specifically is affected.
> 
> The widgets, however, seem to know the following body parts:
> 
> image
> 
> You've seem to have already added these:
> 
> ```
> IflatablePlug_Inflate,
> IflatablePlug_Deflate
> ```
> 
> Some UD-Specific events would be cool like:
> 
> ```
> PlugVaginal_Start, PlugVaginal_Stop
> PluginVaginal_Inflate, PiercingVaginal_Deflate
> 
> PlugAnal_Start, PlugAnal_Stop
> PluginAnal_Inflate, PiercingNipple_Deflate
> 
> PiercingClit_Start, PiercingClit_Stop
> PiercingNipple_Start, PiercingNipple_Stop
> ```
> 
> StrArg could be info about the pattern type, if its on, off etc.. FloatArg could be info about the strength of the vibration
> 
> BTW: Yes, the library I use (bp.io) can control inflators, constrictors (do the opposite of inflators), oscillators, rotators, vibrators, strokers, so the more metadata those events can contain, the merrier.

Nice idea. With UD 3.0, all vibs have new UD_EroZones variable which should contain info about which body part is stimulated (https://github.com/IHateMyKite/UnforgivingDevices/blob/8501cc4480f7da3ee609378c99ca247673c18b6b/Scripts/Source/OrgasmSystem.psc#L54-L67). Will try to cook some events which use this

IHateMyKite:
@kurotatsu77 please close the review. From my understanding you already tested so only closing it should be enough. I want to release next experimental build today, and his is last unmerged thing. Also, in future I will create similar PR for adding new events again, so its not like I will be not adding anything again

kurotatsu77 (approved):
Yep, closing it!

Will try to get my resolve together and actually try to create the quest using these...

garryrot:
That event api looks great, right now doing some refactoring for bigger changes but looking forward integrating that

PR #191, opened 2024-05-11
Added Outfit Manager

IHateMyKite:
- Added new Outfit Manager which will be used instead of existing system for equipping abadon suits

TODO:
- [x] Add MCM page and implement a way to edit the outfit variables (RND and Disable)
- [x] Rework existing suits to use new outfit system (currently only Latex and Simple outfit)


kurotatsu77:
Here's some initial impressions:
1. Currently no way to control selection of outfits themselves. Random checkbox is disabled in both normal and debug modes. Suggestion - replace boolean Random flag with int weight, set it something like 50 by default for all outfits, allow setting for each outfit. When weight set to 0 - outfit is disabled from selection as randomly picked.
2. Abadon plug page with suit selection is not updated, planned for later?
3. A few nitpicks, outfits names cases when selecting them all over the place, better to align them. Also parts weights by default set to strange values, get properly set after pressing Reset command, see screens attached:
ScreenShot13
ScreenShot14

IHateMyKite:
> Here's some initial impressions:
> 
>     1. Currently no way to control selection of outfits themselves. Random checkbox is disabled in both normal and debug modes. Suggestion - replace boolean Random flag with int weight, set it something like 50 by default for all outfits, allow setting for each outfit. When weight set to 0 - outfit is disabled from selection as randomly picked.
> 
>     2. Abadon plug page with suit selection is not updated, planned for later?
> 
>     3. A few nitpicks, outfits names cases when selecting them all over the place, better to align them. Also parts weights by default set to strange values, get properly set after pressing Reset command, see screens attached:
>        ScreenShot13
>        ScreenShot14

1. The random button is only for informing player if the outfit can be randomly locked on. If you want to disable some outfit from being randomly locked, you have to use the Disable button. In essence they do the same thing. Only change is that the Random variable is only used internally and cant be changed. For example, I used it for Abadon Finisher suit so it cant randomly be locked with black goo. Instead, it can only be locked on with script. This can obviously be changed with script, so you can for example change this variable once some quest is finished, or if other criteria is meet
2. Will do it latter. For now it effectively does nothing.
3. I think I found the reason why the Weight values are not correct. Try to save and load the save, that should fix it. I will fix it so its not required. Regarding outfit names, I have no idea why some use different text case. I think this is just some SkyUi bug, as all the names should be in same format. Only SkyUi show them differently

kurotatsu77 (approved):
All in all looks like awesome work, should be appreciated by every user.
All that left after adding this is adding docs on how to create custom outfits for this.

IHateMyKite:
> All in all looks like awesome work, should be appreciated by every user. All that left after adding this is adding docs on how to create custom outfits for this.

Will try to rework the documentation once I have more time. I wanted to add tutorial for adding new modifiers anyway, so I will add this too. For now you can use as reference the Framework Outfit quest

PR #194, opened 2024-05-26
Adding animations (DD NG & UDAnimationsSE)

iiw2012:
Made a registration file of animations from DD NG.

The animation section in MCM has been redesigned and split into two pages: Animations and Animations Playground.
1. The first page containts all the settings, including a list of found and loaded json files.
2. The second page contains the interface for testing animations.

With all these changes I was able to show more information about json load errors and animations received in the request.

Added animations from UDAnimationsSE mod.


IHateMyKite:
Nice work. Great to see you back. Tested it and looks like everything works as intended. Only small point for future: The new DDNG animation file you created checks for DD_Animation_Overhaul_by_Taki17.esp, but this file will be no present in next ddng version, as it is merged in. Also, UD 3.0 will require DDNG to work, so this check might be redundant.

IHateMyKite (approved):

PR #196, opened 2024-06-05
Reworked Bound combat

IHateMyKite:
- Reworked Bound combat by allowing NPC to draw weapons when bound and removing the PreventCombat spell. It is important that latest DDNG is used, otherwise NPCs will use weapons!
- Added fix which prevents npcs from drawing weapons while in animation, as this can break their state


iiw2012:
Maybe it's a problem on my side, but I'm not having animations interrupted when combat starts (animation where the player is helping follower).

When I interrupt the animation manually, the companion just runs around the enemy without trying to hit him.

I'm using EFF.

No errors in Papyrus or UDNative logs (other than the obsolete DD NG).

IHateMyKite:
> Maybe it's a problem on my side, but I'm not having animations interrupted when combat starts (animation where the player is helping follower).
> 
> When I interrupt the animation manually, the companion just runs around the enemy without trying to hit him.
> 
> I'm using EFF.
> 
> No errors in Papyrus or UDNative logs (other than the obsolete DD NG).

For interrupting, are you using the Stop animation from MCM, or stopping the minigame manually? As you will have to stop the minigame, because there are 2 disables placed at ones (1 by minigame, and 1 by animation), and both of them have to be removed. Regarding the minigame stop, the minigame will not end itself when player gets in combat. The reason is that you might sometimes get stuck somewhere where you are constantly in combat, but relatively safe. So you can safely struggle, but are still in combat.

iiw2012:
I did the following experiment:

Ran with a companion to the bandits' camp, and before they realized it, start a help struggle mingame.
A fight started, several arrows hit the player and the companion. Bandits running close to the scene went into a passive state and stopped attacking.

All this time the minigame was not interrupted (it is done on purpose, as I understand).

When I stopped the minigame manually (with the E button), the bandits switched to aggressive state, but the companion did not. The companion ran up to the bandits, but did not try to attack.

This state of companion stopped only when I killed all the enemies myself.

I understand now that this is not related to the PR changes (I've encountered this behavior before) and was not the purpose of the fixes. Some other bug. Possibly even related to the EFF.

IHateMyKite:
> I did the following experiment:
> 
> Ran with a companion to the bandits' camp, and before they realized it, start a help struggle mingame. A fight started, several arrows hit the player and the companion. Bandits running close to the scene went into a passive state and stopped attacking.
> 
> All this time the minigame was not interrupted (it is done on purpose, as I understand).
> 
> When I stopped the minigame manually (with the E button), the bandits switched to aggressive state, but the companion did not. The companion ran up to the bandits, but did not try to attack.
> 
> This state of companion stopped only when I killed all the enemies myself.
> 
> I understand now that this is not related to the PR changes (I've encountered this behavior before) and was not the purpose of the fixes. Some other bug. Possibly even related to the EFF.

I will try to replicate the bug same way. Untill then, did you try a new game ? As I removed implementation of the PreventCombat spell, but if you are using old save, the spell will still be applied, and never removed

IHateMyKite:
Tested it, and it works. Either its caused by not using the latest DDNG, or by the EFF you mentioned. You might try to instead use [NFF](https://www.nexusmods.com/skyrimspecialedition/mods/55653), which I use and it works

iiw2012:
> I will try to replicate the bug same way. Untill then, did you try a new game ? As I removed implementation of the PreventCombat spell, but if you are using old save, the spell will still be applied, and never removed

Yeah. I got lazy and used an old save.

A new game didn't help, and the companion is still running around enemies the same way. 

But other than that, everything works.

iiw2012 (approved):

PR #195, opened 2024-05-28
More modifiers and Abadon Armor

iiw2012:
I prepared a small expansion of the modifier list, and added a couple features to the existing modifiers.

Added new events for processing in modifiers:
1. **GameLoad** - event when loading the game. Called once during UD update.
2. **WeaponHit** - event when player takes damage. I had to add a timeout to the OnHit event to cut off too frequent triggers.
3. **ConditionLoss** - event when the condition of the device gets worse.

Added and changed some modifiers:
1. **Armor Modifier**. When equipping, it adjusts the base form of the device: adds the necessary keywords, enables flags and sets the armor value. I had to add a GameLoad event to restore all those values. Perhaps it would be easier to customize the form in CK, and make this modifier just decorative.
2. "**Manifest on Damage**" modifier. A device with this modifier is able to summon other devices after taking damage. I used the new WeaponHit and ConditionLoss events here.
3. **Evolution Modifier** can now be triggered by damage taken (see WeaponHit and ConditionLoss events above).

The ultimate goal of these changes is to add armor sets that will evolve as you use them. This will diversify the player's experience of interacting with devices that will not only punish, but also help in some way.

So far, I've created a heavy armor set (for now, it's called Abadon Warrior Armor) with progressive evolutions and manifestations.

The following is planned:
1. Add a light armor set (e.g. Abadon Scout Armor).
2. Scatter pieces of armor around the loot table. Maybe add a simple quest.
3. Add a "spicy" ending for users of the added armor sets.
4. Add more modifiers that benefit the player.

How to test:
I've set high values for EVL and DMA modifiers. There is also _DebugDamageMutliplier variable that multiplies all probabilities by 10. And the rechargeable bear traps at Halted Stream Camp are very handy for generating OnHit events. So it won't take long before the protagonist is wearing the entire set of armor, starting with a single item (hopefully any item).

PR #197, opened 2024-07-04
Added  new native functions for modifier handling

IHateMyKite:
- Added new native functions for modifier handling as mentioned here https://github.com/IHateMyKite/UnforgivingDevices/pull/195#issuecomment-2165798749


kurotatsu77 (approved):
As far as I can see - no problems!
Of course, @iiw2012 would be much more qualified to review this, but...

PR #213, opened 2024-12-12
improved version of prevent unequip

naitro2010:


naitro2010:
@ThisIsPatrick Thanks for letting me know.  I'm not very familiar with Papyrus scripting yet.  I'm removing the unused imports in the next commit.  :)

Scripts/Source/UD_CustomDevice_RenderScript.psc
```diff
@@ -1496,12 +1497,20 @@ Function unlockRestrain(bool abForceDestroy = false,bool abWaitForRemove = True)
...
+               removeDevice(Wearer)
                 return
             endif
         endwhile
     else
+        zadNativeFunctions.SetDisableUnequip(Wearer,deviceInventory,false)    
+        zadNativeFunctions.SetDisableUnequip(Wearer,deviceRendered,false)
         libs.UnlockDevice(Wearer, deviceInventory, deviceRendered, UD_DeviceKeyword, zad_DestroyOnRemove || hasModifier("DOR") || abForceDestroy)
+        removeDevice(Wearer)
+        removeDevice(Wearer)
```
IHateMyKite, line 1513: Why is removeDevice called 2 times ?

naitro2010: UnlockDevice sometimes fails to fully remove the RenderDevice.  I added some additional code to removeDevice so if the first removeDevice doesn't fully remove the device the second run will.  Line 7838 of UD_CustomDevice_RenderScript.psc has the changes.

naitro2010: I think it is a separate issue that usually causes console error messages and sometimes prevents render devices from being removed without the pull request changes. It might be a timing bug.  The pull request seems to cause it to happen more frequently possibly because of timing changes.  I don't know how to fix the timing issue or even which script it is in.  The removeDevice change is mostly a workaround until the timing issue can be fixed.  The original pull request was calling the general fix code as a workaround which took a very long time to run so I changed it to call removeDevice instead in case the device isn't already unequipped.

naitro2010: If someone more familiar with Papyrus can figure out where the timing issue could be it would be very helpful.

naitro2010: I might write a native SKSE64 plugin to help debug timing issues by adding a log file for all Papyrus operation start and stop times.

PR #215, opened 2025-02-22
Fixes start event being sent instead of end event

naitro2010:


IHateMyKite:
Thank you for fix. Good work

IHateMyKite (approved):

PR #216, opened 2025-03-04
add fixes for removeAllDevices and removeUnusedDevices

naitro2010:
Fixes removeAllDevices getting stuck in a loop.
Fixes widgets not being cleared for unused devices.

### DIFF
--- a/Scripts/Source/UD_CustomDevice_NPCSlot.psc
+++ b/Scripts/Source/UD_CustomDevice_NPCSlot.psc
@@ -847,6 +847,14 @@ Function removeAllDevices()
     ;startDeviceManipulation()
     StorageUtil.SetIntValue(getActor(), "UD_blockSlotUpdate",1)
     while UD_equipedCustomDevices[0]
+        if UD_equipedCustomDevices[0].isUnlocked
+            Utility.waitMenuMode(0.2)
+            if UD_equipedCustomDevices[0].isUnlocked
+                removeCopies()
+                removeUnusedDevices()
+                removeLostRenderDevices()
+            endif
+        endif    
         UD_equipedCustomDevices[0].unlockRestrain()
     endwhile
     ;regainDevices()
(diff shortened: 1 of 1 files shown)


Scripts/Source/UD_CustomDevice_NPCSlot.psc
```diff
@@ -847,6 +847,14 @@ Function removeAllDevices()
     ;startDeviceManipulation()
     StorageUtil.SetIntValue(getActor(), "UD_blockSlotUpdate",1)
     while UD_equipedCustomDevices[0]
+        if UD_equipedCustomDevices[0].isUnlocked
```
IHateMyKite, line 850: Should not the condition be after the unlockRestrain is called first ? As otherwise the condition will always be true for the first time

Scripts/Source/UD_CustomDevice_NPCSlot.psc
```diff
@@ -867,6 +875,9 @@ Function removeUnusedDevices()
                 loc_device.removeDevice(loc_device.getWearer())
                 loc_device.getWearer().removeItem(loc_device.deviceRendered)
             endif
+            UDmain.Log(loc_device.getDeviceName() + " clearing widgets")
```
IHateMyKite, line 878: Please either remove the logs or add "if UDmain.TraceAllowed()". Or use UDMain.Info, which should be used if the inforamation should be shown to user in console

IHateMyKite (changes requested):

Scripts/Source/UD_CustomDevice_NPCSlot.psc
naitro2010: unlockRestrain sets isUnlocked before it is supposed to remove it from UD_equipedCustomDevices.  I added the check above so it only runs if it didn't fully remove it from UD_equipedCustomDevices and it already ran the unlockRestrain on the same slot.

Scripts/Source/UD_CustomDevice_NPCSlot.psc
naitro2010: I'm removing the extra logs now.

IHateMyKite (approved):

IHateMyKite:
Thank you, good job

Issue #217, opened 2025-03-11
Struggle crits not working with gamepad


elliesec:
**Describe the bug**
When playing with a gamepad, crits during the struggle minigame either do not work at all, or always fail. If I rebind my key bindings to the keyboard, crits work as expected (and I can trigger them reliably).

**To Reproduce**
1. Configure stamina/magicka hotkeys to gamepad buttons via the UD MCM
2. Equip a device
3. Start the regular struggle minigame, or the lockpicking minigame (I've observed this behaviour in both minigames, but the orgasm resist minigame works correctly)
4. Wait for a crit and press the corresponding button
5. Nothing happens, or a crit failure is always triggered (see additional notes below)

**Expected behavior**
Pressing the correct crit button within the permitted time should result in a successful crit being triggered.

**Environment**
* Game version: AE 1.6.1170
* OS: Windows 11
* Gamepad: XBox One controller (plugged in, not wireless, so I wouldn't expect latency to be an issue) - haven't tested with other gamepads, but I can if needed
* UD version: nightly build 20250309 (https://github.com/IHateMyKite/UnforgivingDevices/commit/27e5c5fcfd9748cc40a199568c08e97d82de62f3), although AFAIK this has been around for a while (I've definitely noticed it happening on at least a few of the more recent experimental builds posted in the DD Discord server)
* DDNG version: 0.4

**Additional context**:
* Auto crit setting is off, and the mandatory crits setting is off
* I have maxed out the crit duration adjustment in the MCM, to try to ensure I'm not just missing the crits
* For some buttons (RT/LT/RB/LB), nothing happens when the button is pressed (i.e. no crit success or failure - it just behaves as if I've missed the crit), but for others (A/B/X/Y), pressing the button consistently results in a crit failure. I've not tried other buttons.
* Notably, crits are working correctly with the gamepad in the orgasm resist minigame (which at a glance looks like it's doing its key handling via Papyrus rather than a UD native callback)
* If I rebind my keys to the keyboard, crits work as expected
* I've definitely had crits working fine with the gamepad on a previous version of UD, although I can't remember which (possibly the v2 published on LL)

PR #218, opened 2025-03-11
Fixed gamepad controls

IHateMyKite:
Fix issue #217 

Most changes are in native plugin. Only small adjustment was done in papyrus. See https://github.com/IHateMyKite/UnforgivingDevicesNative/commit/872c8813292f8fcf435aa746cbc5c5ad3d2db7ae


kurotatsu77:
Currently testing this PR, spotted quite a number of problems with the new modifiers, not sure if this one introduces them:
1. Temporary struggle disable is not working. It stays blocked even after 50+ hours of being worn.
2. Devices which has to be placated with stolen sweet roll do not detect one stolen sweet roll present in inventory. Generally I feel like this design choice is not the best. I'd propose to make trigger into placating devices with black goo and concentrated black goo, dropping the whole stolen thing completely.
3. Currently all devices can get restraints manifest effect, is this intentional? 
4. When wearing some devices, with temporary disabled struggling, game has a lot more 1 second stutters than usual.

iiw2012:
> Currently testing this PR, spotted quite a number of problems with the new modifiers, not sure if this one introduces them:

It's probably my fault and not this PR.

> 1. Temporary struggle disable is not working. It stays blocked even after 50+ hours of being worn.

I've noticed so far that timed shields on the locks get closed forever. Still looking for the source of this problem.

> 2. Devices which has to be placated with stolen sweet roll do not detect one stolen sweet roll present in inventory. Generally I feel like this design choice is not the best. I'd propose to make trigger into placating devices with black goo and concentrated black goo, dropping the whole stolen thing completely.

But it's a meme modifier :(

Jokes aside, the ItemObtain and ItemDemand triggers are really crooked, and I haven't figured out how to do them properly yet.

> 3. Currently all devices can get restraints manifest effect, is this intentional?

Manifest modifiers have two presets each, one of which will allow them to appear on any device. If that makes the devices too complicated, I can tweak the probabilities.

> 4. When wearing some devices, with temporary disabled struggling, game has a lot more 1 second stutters than usual.

I'll look into it.

I'll do a separate PR with fixes when I have something to show.

kurotatsu77:
OK, this has to be some kind of aliases/pointers/something else mismatch:
I've had on player "Tight fashionable corset" from deviously accessible, it's patched using Synthesis. It's not worn, just present in inventory. When got follower (Chloe, from DCL, most probably not important), I've spotted that after anything happening like adding, removing devices on player or Chloe, in Chloe's device ist started appearing that corset. After running fixes it disappears, only to appear some time later. After I've moved it into Chloe's inventory (again, not worn, simply stored), that behavior stopped.

On unrelated note - looks like English translation bug, in the UD MCM Debug menu label for display details is not working correctly, showing translation tag.

kurotatsu77 (approved):

PR #221, opened 2025-03-21
More Modifiers Fixes

iiw2012:
- Changed algorithm of adding modifiers in the patcher. Made the results more predictable.
    - A modifier with an absolute probability of 100% will ALWAYS appear on suitable devices.
    - A modifier with a weighted probability of 50% will appear 2 times less often than a modifier with a weighted probability of 100%.
- BlockMinigame outcome.
    - Fixed errors.
    - Added messages when the state has changed.
- ItemObtain trigger. 
    - Fixed errors.
    - Improved detection of stolen items.
- Reduced the probability of MAN modifiers (OLD and TLD) for generic devices (5% weighted probability). 
- Turned off the NST modifier (0% prob.) until I figure out its item filter.


kurotatsu77:
Another thing spotted, might be good to have a look at: lockpicking trigger works fine when picking locks on doors and chests, but seem to ignore picking locks on restraints.

iiw2012:
> Another thing spotted, might be good to have a look at: lockpicking trigger works fine when picking locks on doors and chests, but seem to ignore picking locks on restraints.

When I tested this trigger a couple days ago, I noticed that it was responding to locks on the device. And was just wondering whether to silence such events. 
So it's strange that you don't have those events already :)

I'll look into it and see what's going on.

IHateMyKite:
> * Changed algorithm of adding modifiers in the patcher. Made the results more predictable.
>       
>       * A modifier with an absolute probability of 100% will ALWAYS appear on suitable devices.
>       * A modifier with a weighted probability of 50% will appear 2 times less often than a modifier with a weighted probability of 100%.
> 
>     * BlockMinigame outcome.
>       
>       * Fixed errors.
>       * Added messages when the state has changed.
> 
>     * ItemObtain trigger.
>       
>       * Fixed errors.
>       * Improved detection of stolen items.
> 
>     * Reduced the probability of MAN modifiers (OLD and TLD) for generic devices (5% weighted probability).
> 
>     * Turned off the NST modifier (0% prob.) until I figure out its item filter.

I think it will be best to just change chance of MAN modifiers for generic devices to 0%. Then users can change it in MCM. This way no one will complain about it getting on every device (even when chance will be smaller)

iiw2012:
> I bet it should react to the locks of devices being picked! I wonder if difference of reaction can be tied to the changes in the latest PR IHMK merged now. Try with it integrated?

Checked the trigger with the latest changes, still works. Since I'm using tracked stat, it's hard for me to imagine what could make events register differently. Perhaps the game version or SKSE core dll? I've checked on 1.6.1170.


> I think it will be best to just change chance of MAN modifiers for generic devices to 0%. Then users can change it in MCM. This way no one will complain about it getting on every device (even when chance will be smaller)

Then there's probably only one person who will see it - me.
I still think we should leave room for the unexpected. So far I've changed the preset settings to remove repeats and make it easier.

I think the main problem is not even the events, but the fact that it's hard for the player to navigate through a large number of different types of triggers. It's a pain to go through all the devices to understand the possible “threats” and prioritize them.

I'll think of a way to make it easier.

kurotatsu77:
Report from Discord : Exploration modifier doesn't gets triggered, despite discovering a lot of new locations, I can confirm it.

Regarding lockpicking - do you use Experience and Ordinator?

Regarding manifest modifiers - previously they were restricted to plugs and piercings only for patched devices. Maybe add checkbox "Plugs and piercings only" into those, checked by default?

iiw2012:
> Report from Discord : Exploration modifier doesn't gets triggered, despite discovering a lot of new locations, I can confirm it.

I fixed it among the first edits. The trigger itself worked fine (tracked stat "Locations Discovered"), the error was in the outcome BlockMinigame, in conjunction with which it was used in a single modifier EXLP.

> Regarding lockpicking - do you use Experience and Ordinator?

Experience v3.5. Still don't understand how it could affect game tracked stats.

> Maybe add checkbox "Plugs and piercings only" into those, checked by default?

Since I redesigned the way modifiers are set up for Patcher, it would be inconvenient to add such a setting.

To avoid confusion, we're talking about the new OLD & TLD modifiers, which inherit the functions of the old MAN modifier and have tag MAN.

Now there is a separate settings preset for sex toys (almost like in the original). And I added a preset for all other devices with a much lower probability of occurrence and more gentle difficulty settings. The question is whether to disable it entirely, or adjust it further.

iiw2012:
If the appearance of rare modifiers has become a problem, then I may have broken something in the Patcher algorithm. I'll check it out.

iiw2012:
> I'll think of a way to make it easier.

I've created this device list as a prototype (cloned from the uiextensions listmenu).

UD_ListMenu_2_75

The device name, health and modifier tags are displayed in one place. I hope this makes it easier to prioritize devices. 

I'll make a separate PR when I make this list functional. If there are any suggestions, I can try to implement them in this concept.

kurotatsu77:
Found one more small thing - punisher piercing has MAO with repeat set to false, I think it's against it's intended design.

iiw2012:
I think we can shove this PR into the main to at least apply some critical fixes.

Also some balance changes:
- Severely reduced the number of modifiers on items. 0 - 2 on average.
- Hardcore modifiers appear less frequently.

For now, I'll look for the remaining bugs and try to come up with more simple modifiers.

PR #226, opened 2025-04-18
Small prerelease fixes

IHateMyKite:
Fixes to small issues found during testing

### DIFF
--- a/Scripts/Source/UD_MCM_script.psc
+++ b/Scripts/Source/UD_MCM_script.psc
@@ -2186,6 +2186,7 @@ Function OnOptionSelectUI(int option)
         forcePageReset()
     ElseIf option == UD_WidgetTest_T
         closeMCM()
+        Utility.Wait(0.5)
         UDWC.TestWidgets()
     ElseIf option == UD_WidgetReset_T
         closeMCM()
--- a/Scripts/Source/UD_WidgetControl.psc
+++ b/Scripts/Source/UD_WidgetControl.psc
@@ -1310,6 +1310,7 @@ EndFunction
 Bool _InitWidgetMutex = False
 Bool _InitMetersMutex = False
 Bool _OnUpdateMutex = False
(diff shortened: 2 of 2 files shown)


Scripts/Source/UDCustomDeviceMain.psc
```diff
@@ -1730,6 +1728,62 @@ Bool Function ShowUndressArmorMenu(Actor akActor)
...
+        UDmain.Print("You dont have any ideas yet. You should wait some more time")
+        return
+    endif
+    
+    float loc_cooldownHours = 0.0
+    
+    ; Currently only one option
+    if UDmain.Player.getItemCount(Lockpick) <= 0
+        UDmain.Player.AddItem(Lockpick,RandomInt(5,10),true)
+        UDmain.Print("You improvised and created some makeshift lockpicks from things you could find around")
```
iiw2012, line 1777: During my tests I noticed something in the UI, not directly related to the changes, but still.

When these messages are displayed, they go to the standard notification place (upper-left conner). Because of the "IsAnyMenuOpen" condition in UD_WidgetControl. I've turned that condition off, and it seems that iWW is able to display  messages in the center of the screen at the same time as the visible UserList menu.

https://github.com/IHateMyKite/UnforgivingDevices/blob/19a90d523f67796188036e8020a897e074da9958/Scripts/Source/UD_WidgetControl.psc#L1809-L1813

I suggest making an exception in this condition at least for the CustomMenu. Assuming I'm not imagining it, and it works with custom menus.

iiw2012 (approved):

Scripts/Source/UDCustomDeviceMain.psc
IHateMyKite: Think I got it. Basically reworked it so it checks all menus except the custom menu.
Can you test it one more time ?

kurotatsu77 (approved):
Looks good to me.

PR #235, opened 2025-05-18
Fix BRRF compatibility. Removed error log

### LINKED ISSUE #233: Abadon Devices comaptibility with Beast races
**Describe the bug**
Some Abadon gags are not working with elfs

**To Reproduce**
Equip Abadon Gag on elf

**Expected behavior**
The gags should show

**Screenshots**
n.r

n.r

**Additional context**
[Report](https://www.loverslab.com/topic/250876-unforgiving-devices/page/2/#findComment-7033898)

### LINKED ISSUE #234: [UD,!ERROR!,T=347.839020]: Cant set lastopeneddevice  on pressing device key
**Describe the bug**
A lot of users report this error, which pops up when no heavy bondage worn and no other devices were struggled against previously.

**Expected behavior**
Possible solution: when lastopeneddevice can't be set - open device list instead.

IHateMyKite:
- Fix #233 
- Fix #234 
- Additional fixes


kurotatsu77 (approved):

Scripts/Source/UD_CustomDevice_NPCSlot.psc
```diff
@@ -889,7 +889,8 @@ int Function unregisterDevice(UD_CustomDevice_RenderScript oref,int i = 0,bool s
             UD_equipedCustomDevices[i] = None
         Else
         ; ???
-            UDmain.Warning(Self + "::unregisterDevice() Something wrong with UD_equipedCustomDevices array. Unexpected element value.")
+            ; This is intended behaviour when device is not first element in array and i = 0. Why print warning ???
```
iiw2012, line 892: It was assumed that hitting this branch was only possible if we found None in the array before we deleted the element we were looking for. Such a case is possible if the array has empty elements in the middle, or that the element we are looking for does not exist.

In this case, I force sorting (alignment) of the array and write it to the log.

iiw2012: I can see now that the condition is incorrect and includes extra cases.

It should be something like this:
```papyrus
        ElseIf res == 0 && UD_equipedCustomDevices[i] == None
            loc_sort = True
        endif
```
Or remove this as it is, since these cases are probably impossible.

IHateMyKite: For now I will remove it, as I'm worried that the additional case will break something else. Having empty slot in middle should be impossible (unless user unregister the device manually somehow). Every time device is unregistered, the array is also sorted.

Issue #237, opened 2025-05-21
Add messages about modifiers actions


iiw2012:
**Is your feature request related to a problem? Please describe.**
Modifiers can be confusing and more information are needed about its nature and possible actions.

**Describe the solution you'd like**
Standalone modifiers and ModOutcome scripts should print some notifications about its actions.

**Describe alternatives you've considered**
Add visual cues if possible.

**Additional context**
[Forum message](https://www.loverslab.com/topic/250876-unforgiving-devices/page/5/#findComment-7037181)

Issue #236, opened 2025-05-19
Probabilities in presets (first preset) for modifiers don't work as expected


iiw2012:
**Describe the bug**
Modifiers with a set 100% abs. probability are not applied to devices.

**To Reproduce**
Firts discovered on TMREG modifier.
1. Open settings for Generic preset of TMREG modifier in MCM.
2. Change probability to 100% and toggle "Is Absolute".
3. Equip any suitable device (non-ebony).
4. Device will be without the TMREG modifier

**Expected behavior**
Device must receive the TMREG modifier

**Additional context**
[Report on forum](https://www.loverslab.com/topic/250876-unforgiving-devices/page/4/#findComment-7035713)

PR #228, opened 2025-05-07
Updated Dialogues. Some other possible fixes

IHateMyKite:
Fix #227
Updated Dialogues, so children cant be valid targets. Fixed one of the dialogue options which was not working till now

PR #242, opened 2025-05-28
Fixes for modifiers, widgets and more

### LINKED ISSUE #239: Overlapping states in WidgetControl when the quest resets (OnUpdate call delay after Reset UI)
**Describe the bug**
When base quest is reset, Widget Control script suffers from states overlap and function calls from different state blocks.

It is not a bug that could be fixed in UD. I've described this bug so we don't forget it exists. It doesn't actually break anything, but I'll try to get around it.

**To Reproduce**
1. Use Reset UI button in MCM
2. Wait for it to end.
3. See in log that quest OnUpdate function is called in 30 second instead of "Animation_Update" period.

**Expected behavior**
OnUpdate function should be called almost immediately to update all widgets.

 UD version 3.0

**Additional context**
Here is the log:
```
[UD,1,T=30.017002]: Print -> UI reset start - wait!
[UD,3,T=30.304001]: UD_WidgetControl::OnInit() State = ; RegisterForSingleUpdate(6)
[UD,3,T=36.330002]: UD_WidgetControl::OnUIReload() abGameLoad = False
...
[UD,1,T=36.471001]: Print -> UI reset done!
[UD,3,T=36.522003]: UD_WidgetControl::OnBeginState() State = iWidgetInstalled
[UD,3,T=36.543003]: UD_WidgetControl::OnBeginState() State = iWidgetInstalled; RegisterForSingleUpdate(_Animation_Update)
...
[UD,3,T=36.647003]: UD_WidgetControl::OnUpdate() State = iWidgetInstalled; RegisterForSingleUpdate(30)
```
The most curious thing here is the last line. The OnUpdate function was called from a block with no state (long update period), but GetState returned the iWidgetInstalled state.

iiw2012:
Accumulated changes from tests and forum posts

Improvements in WidgetControl:
- Removed redundant requests for widget registration.
- Fixed icon transparency animation (previously it could hang on a random value when overlapping calls).
- Saving and restoring LinkActor for the orgasm widget (don't know if this is necessary).
- Fixed #239

Improvements to modifiers:
- #237
- Restored missing ENCH1 and LCOL modifiers.
- A bunch of bug fixes and parameter customization.
- Added new tags: ALCO, CRIME.
- TimeHour trigger fires at different times for different devices and modifiers.
- Fixed somehow #236

Interface:
- Added hints in the Lock window of the device menu if the character doesn't have lockpicks and/or the right key.


iiw2012:
I forgot to mention.

Restored default values for StruggleCrit properties in CustomDevice_RenderScript. Perhaps this will solve at least some of the problems with crits on devices from the forum.

IHateMyKite:
The orgasm linking is not persistent. It have to be called every time game is reloaded. Did you have the issue where the meter didnt work and if yes did this solve the issue ? As I cant replicate the issue and never had it for some reason


> I forgot to mention.
> 
> Restored default values for StruggleCrit properties in CustomDevice_RenderScript. Perhaps this will solve at least some of the problems with crits on devices from the forum.

Seems to be the source of issue. I possibly forget that some Abadon devices do not have this variables set.

iiw2012:
> The orgasm linking is not persistent. It have to be called every time game is reloaded. Did you have the issue where the meter didnt work and if yes did this solve the issue ? As I cant replicate the issue and never had it for some reason

No, orgasm bar does not work and nothing helps. If I run the Test UI, the bar fills for the duration of the test.

Sometimes I notice a couple of pink pixels on the edge, so I can assume there are problems with the scale of the bar. But I can't say for sure.

I'll try some more tests.

iiw2012:
> If it will work, it means that issue was introduced with PR https://github.com/IHateMyKite/UnforgivingDevices/pull/226

Yes, it works better:
- After starting the game - works fine.
- After loading the game - works fine.
- After changing position - the filling disappears.
- After Reset UI - the filling disappears.


I've also checked the iWW NG log on the latest version, it shows errors and warnings when loading the game, Reset UI or moving widgets around:
```
[log] [error] [IWWFunctions.cpp:669] Failed getting hud - loc_hudmenu => null
[log] [warning] [IWWFunctions.cpp:458] Invalid id passed (-1) => skipping function call
[log] [warning] [IWWFunctions.cpp:458] Invalid id passed (-1) => skipping function call
[log] [warning] [IWWFunctions.cpp:458] Invalid id passed (-1) => skipping function call
[log] [warning] [IWWFunctions.cpp:458] Invalid id passed (-1) => skipping function call
[log] [warning] [IWWFunctions.cpp:458] Invalid id passed (-1) => skipping function call
[log] [warning] [IWWFunctions.cpp:458] Invalid id passed (-1) => skipping function call
[log] [warning] [IWWFunctions.cpp:458] Invalid id passed (-1) => skipping function call
```

IHateMyKite:
> > If it will work, it means that issue was introduced with PR #226
> 
> Yes, it works better:
> 
>     * After starting the game - works fine.
> 
>     * After loading the game - works fine.
> 
>     * After changing position - the filling disappears. Loading a save helps.
> 
>     * After Reset UI - the filling disappears. Loading a save helps.
> 
> 
> I've also checked the iWW NG log on the latest version, it shows errors and warnings when loading the game, Reset UI or moving widgets around:
> 
> ```
> [log] [error] [IWWFunctions.cpp:669] Failed getting hud - loc_hudmenu => null
> [log] [warning] [IWWFunctions.cpp:458] Invalid id passed (-1) => skipping function call
> [log] [warning] [IWWFunctions.cpp:458] Invalid id passed (-1) => skipping function call
> [log] [warning] [IWWFunctions.cpp:458] Invalid id passed (-1) => skipping function call
> [log] [warning] [IWWFunctions.cpp:458] Invalid id passed (-1) => skipping function call
> [log] [warning] [IWWFunctions.cpp:458] Invalid id passed (-1) => skipping function call
> [log] [warning] [IWWFunctions.cpp:458] Invalid id passed (-1) => skipping function call
> [log] [warning] [IWWFunctions.cpp:458] Invalid id passed (-1) => skipping function call
> ```

Checked my log and I have exact same lines. So this cant be the source of issue. Can you confirm if this lines are also present with the previously mentioned NB ?

Also, I said it wrong. The changes from #226 work, and after that it does not work (next PR is last one before release, so its basically what is in 3.0)
So the actual source of issue is PR #228. Can this be possible source of issue ? https://github.com/IHateMyKite/UnforgivingDevices/blob/23346424d40eb750e1827a20b4ae71659472bf16/Scripts/Source/UD_WidgetControl.psc#L1352

iiw2012:
> Also, I said it wrong. The changes from https://github.com/IHateMyKite/UnforgivingDevices/pull/226 work, and after that it does not work (next PR is last one before release, so its basically what is in 3.0)

I've tested two builds:
1. Build-218 https://github.com/IHateMyKite/UnforgivingDevices/actions/runs/14882318777
2. Build-220 https://github.com/IHateMyKite/UnforgivingDevices/actions/runs/15038689300

The first build works fine with orgasm bar. On the second build it is broken.
I've loaded save from Build-218 on Build-220 and orgasm bar didn't work.
I've loaded save from Build-220 on Build-218 and orgasm bar worked.

I tested this several times and these results were always repeated. The problem must have appeared somewhere between these builds.

> Checked my log and I have exact same lines. So this cant be the source of issue. Can you confirm if this lines are also present with the previously mentioned NB ?

I correlated the occurrence of errors with actions in the game and this is what I got.
```[log] [error] [IWWFunctions.cpp:669] Failed getting hud - loc_hudmenu => null```
Appears at the very beginning of loading a save on both builds. A couple of times I didn't see this line, but I can't tell what caused it.


```[log] [warning] [IWWFunctions.cpp:458] Invalid id passed (-1) => skipping function call```
Does not appear on Build-218. Or very rarely in 1 instance, if you pull all knobs and buttons in MCM at once.
Appears on Build-220 every time OnUIReload is called. 3 lines at once.

> Can this be possible source of issue ?

I removed the new condition on the last build, recompiled the script, but I still didn't have the orgasm bar working for me.

I'll try to figure out the difference between Build-218 and Build-220 for now.

iiw2012:
Well, I reduced the difference to the UD_WidgetControl.pex file. But I'm having trouble figuring out what specific changes break everything. Sometimes one thing works for me, sometimes another. Sometimes on one save, sometimes on another.

We can roll everything back to UD_WidgetControl.pex from Build-218, and it seems to work consistently. But there's clearly a problem with initialization and the call race somewhere, which I want to unwind after all.

iiw2012:
So far I have found two sources of problems:

1. When initializing the module there is unpredictable switching between script states, unnecessary creation/destruction of widgets. I solved this by adding explicit script state checks to avoid redundant transitions.
2. `UD_Native::LinkActorToMeter()` requires a widget id, but this id appears relatively late when all widgets are created. I'm now trying to implement delayed (asynchronous) initialization.

But something else is getting in the way. On a new game or saves before 3.0 everything works. But when I reload a save, something breaks again.

I have a suspicion that the issue is a `UD_Native::AddMeterEntryIWW()` call for "player-orgasm" with zeroes. I'don't fully understand what it does, but everything seems to work better without it.

iiw2012:
> * AddMeterEntryIWW - Makes native plugin to update the meter on every frame. This is also used for struggle bars, and they work so that should be not the issue
> * LinkActorToMeter - Tells native plugin that the meter should be updated on every frame with value of orgasm progress from passed actor. If this is not called, or called incorrectly,  the meter will not update. Which is what is happening.

I see. Do these functions save any information when saving the game or relay on any previously saved data? Can I assume they behave the same when starting a new game, loading a game, loading an older version of a game?

iiw2012:
Tested the latest changes and almost everything seems to work. 
There is still a problem though. When switching the position of meters (left, center, right) they can lose alignment or move away from the desired position. This is an old problem, and seems to only appear at screen resolutions of 21:9 or higher. I used to solve it by recreating the widgets when the position changed. And I've removed that now as a potential source of problems.

In any case, when saving and loading a save, the positioning problems should be solved.

IHateMyKite:
Nice job. Closing this. I will share it on discord for testing before I upload it on LL, and then I will release next UD version

IHateMyKite (approved):

PR #252, opened 2025-08-07
Fixes and adjustments

IHateMyKite:
- Rebalanced manifest modifier. 
- Added reset modifiers button to MCM. 
- Fixed power struggle strength not to update correctly according to device remaining health


kurotatsu77 (approved):
Works as it should for me, great work!

My suggestion would be to add the following in the changelog:
"If you upgrade UD to this version on running save please do not forget to go to Skills tab in the MCM and properly set parameters there as effect of values was changed tremendously."

IHateMyKite:
> Works as it should for me, great work!
> 
> My suggestion would be to add the following in the changelog: "If you upgrade UD to this version on running save please do not forget to go to Skills tab in the MCM and properly set parameters there as effect of values was changed tremendously."

Will do. Thanks for the tip

PR #254, opened 2025-08-24
Just one annoying typo, that keeps popping up again and again in game…

johnsonjoe1:
Just one tiny but annoying typo, that keeps popping up again and again in game.  Makes me wanna fix it.  Thank you for considering.


IHateMyKite (approved):
Thank you for contribution. Merged

PR #258, opened 2025-09-28
Added some more verbosity to the simple you-are-busy or you-are-doing-something message

johnsonjoe1:
Added some more verbosity to the simple you-are-busy or you-are-doing-something message.

This is because people in the forum ask about the reason why they get the busy message permanently.  In many cases the reason comes from another mod than UD, when another mod didn't end their scenes properly.  (A good example is The Forbidden Tome, which always does that to me.)  
Then they come to the UD forum for support, and the information they have is minimal.
So this will help to debug and state more information without breaking immersion I hope.

Sorry for the mess with this "merge commit".  It's only the second or third time I use git.


IHateMyKite (approved):
Nice work. Thank you.
As you said, this should make it more clear for user who encounter this issue. I personally encountered this issue only once when doing LBS quest, so it didnt occur to me that others might not be that lucky

